PR #18, opened 2019-10-30
IEC 61131 reference repository name changed

ntphx:
Changed name of the IEC-61131-3 repository to IEC-61131 and link reference to project landing page instead of `st.md` file. IEC-61131-3 in fact refers to part 3 of the standard (PLC Programming Languages) but the reference document is growing outside of that scope by looking at PLC system design and CODESYS implementations.


MichielVanwelsenaere:
merged! thank you for your contribution!

PR #176, opened 2026-08-14
Sync installation projects from the reference, and replace Invert with RelayType

MichielVanwelsenaere:
Two related pieces of work on the same theme: making a *reference* project and a
building's *installation* project maintainable from one place.

## 1. `sync-implementation-project`

A skill that brings a real building's project up to this one's function blocks
while leaving its logic alone, plus the harness changes that make that possible:

| | |
|:--|:--|
| `-Project <path>` | point every task at another project |
| `info` | read-only inventory — IDE version, libraries, devices, per-object hashes |
| `compare` | structural diff between two projects |
| `path` on an edit | disambiguate a POU name (see below) |

Version compatibility is answered in two parts, because the compiler version a
project stores is **not** reachable from the scripting API — a reflection sweep
over the application object finds nothing, so don't go looking again. Instead:
`info` reports the IDE version, library set and controller, and `verify -Baseline`
builds the thing with the installed compiler. The second is the real gate.

## 2. `RelayType`, replacing `Invert`

`Invert : BOOL` named the implementation — "swap the two payloads" — rather than
the thing being described:

```iecst
FB_DO_SW_RE : FB_OUTPUT_BINARY_MQTT := (FriendlyName := 'Living sockets',
                                        EntityType   := E_MQTT_ENTITY.Switch,
                                        RelayType    := E_RELAY_TYPE.NC);
```

`E_RELAY_TYPE` is a new `qualified_only`, `strict` enum: `NO = 0`, `NC = 1`.
`NO` is the default *and* zero, so an instance that says nothing behaves exactly
as before and no existing project changes meaning.

Renamed in all three places at once — the input on `FB_MQTT_BASE`, both
self-wiring prologues, and all 11 `InitMqttDiscoveryAs*` methods, parameter and
body. Leaving `Invert` on the methods while the declaration said `RelayType`
would be two names for one idea.

NC is not exotic: it is how anything that must survive a dead controller is
wired, sockets above all. Left unsaid, Home Assistant reports such an output
exactly backwards — "off" while the sockets are live — and no build catches it,
because nothing in the PLC is wrong.

## Findings that cost real time

Each of these is written into `CLAUDE.md` or the `codesys-loop` skill, because
none is discoverable from the error message.

**A stale precompile cache fakes a build failure.** `verify` copies the project
to `.ai/work` and builds it alone in an empty folder; `apply` builds the real
project *in place*, beside `<name>_project.precompilecache`. That cache can serve
a stale compiled interface for a block the same run just re-declared:

```
'Invert' is no valid assignment target     x3
Identifier 'Invert' not defined            x3
```

…in a project whose only remaining `Invert` was inside a comment. `do_apply` now
clears the cache and rebuilds once when the first build fails, logging
`precompile cache cleared, rebuilding once`.

Three hours went into the wrong theory first — that a structured initialiser is
stored outside the declaration text like an `FB_init` argument list — because the
same six errors survived `replace_in_decl` per line, a whole-declaration
`decl_replace`, **and** a `decl_replace` that deleted the initialiser outright.
What settled it was `verify -Baseline` on the untouched project coming back
clean. *A project that builds clean untouched but fails an edit-free `apply` is
telling you the difference is the folder, not the edit.*

**A name stops identifying a POU as soon as a project has two controllers.** An
installation with two PFCs has two `PLC_PRG_MAIN`, both real and both owning
text, so every edit addressed by name alone is refused as ambiguous. `"path":
"Wago_G1_Bijbouw/"` picks one. The refusal is the useful behaviour — the
alternative is a coin flip over which building gets rewritten.

**PowerShell variable names are case-insensitive**, so `$project = $referenceProject`
silently overwrote the `-Project` parameter. Every task ran against the reference
project while the banner cheerfully said "override".

**Programs appear twice.** A program exists once as the POU and once as a
task-call node (type GUID `413e2a7d`) carrying no code. Naively indexing by name
kept the empty one, and every program read as "no declaration, no kind".

**`FriendlyName` is declared once on `FB_MQTT_BASE` and inherited**, so looking
only at a derived block's own declaration finds nothing and a migration concludes
there is nothing to do. Walk `EXTENDS`.

**A PLCopen `<dataType>` accepts `<baseType>` then `<addData>`, and nothing
else.** Adding `<documentation>` fails the import with an error naming
`addData` — which reads like an ordering problem and is not. Worse, the import
error is easy to miss: the build then produces a hundred `Identifier not defined`
for the type, and the first message is the real one.

## Scope

Nothing here downloads to any PLC, and no building's project is in this PR —
those live in their own repository.


.claude/skills/sync-implementation-project/SKILL.md
```diff
@@ -0,0 +1,510 @@
...
+
+Run it **after** the block sync, against the post-sync inventory — the
+self-wiring prologue only exists in the project once the blocks have landed.
+Then verify and apply it as a second pass, exactly like §6b but with no
+candidates.
+
+Three things it gets right that a careless hand edit does not:
+
+- **The friendly name comes from the `Name :=` argument, never from the
+  trailing comment.** In Brugsestraat153, `FB_DO_LIGHT_005` is commented
```
MichielVanwelsenaere: Remove this, personal details. Scan entire PR to make sure the project name (= my adres is not mentioned.)

MichielVanwelsenaere: Done, and it was in more places than the line you flagged.

**Removed from the working tree** — 16 occurrences of the project name across 5 files (`sync-implementation-project/SKILL.md`, its `Working-Copy.ps1` and `plan_sync.py`, `codesys-loop/SKILL.md`, `CLAUDE.md`), plus the device name and the room/animal names in the worked example. `SiteA`, `Wago_G1_Annex` and English room names now. The examples lose nothing: the `FB_DO_LIGHT_005` one still shows a trailing comment disagreeing with the announced name, and the pushbutton one still carries a deliberate typo that must be copied verbatim.

**Rewrote the branch.** The name was also in commit 57e9616's message and in the file contents of *both* commits, so a follow-up commit would not have helped — the per-commit diffs still exposed it. The branch is now a single commit `53ec00e`, force-pushed. The full PR diff greps clean for the project name, the device name and every room name.

**Added a guard** in the worked example, where the temptation to paste a real one is highest:

> `SiteA` is a placeholder, and so is every room and device name in this file. An installation project is named after the building it runs, and a building is somebody's address […] This repository is public.

**Also cleared it from my own notes**, which had it too.

The reference project binary and `PLCopen.xml` never contained it — checked.

PR #185, opened 2026-08-18
Add a DFRobot SEN0492 rangefinder block, and a way to reach a device on the wrong baud rate

MichielVanwelsenaere:
Two new blocks and one new capability, **reading on the bench**: distance tracking with a valid measurement state, announced to Home Assistant as a DFRobot device.

The sensor also turned out to have a hard limit on how fast it can be polled, and finding that took most of the work. That story is at the bottom.

```
FB_RS485_SEN0492_1 : FB_RS485_DFROBOT_SEN0492_MQTT(16#50, T#10S)
                   := (FriendlyName := 'Cistern level');
```

That is the whole configuration. Register it on the bus, call it cyclically, and it wires itself from `MqttVariables` and appears in Home Assistant as a DFRobot device with a distance sensor under it.

## The block

`FB_RS485_DFROBOT_SEN0492_MQTT` reads distance and measurement state out of **one** request — they are adjacent holding registers (`16#34`, `16#35`), so the two always describe the same measurement rather than two moments a poll apart.

Function 3 and 6 throughout, not the input registers the Eastron meters use: everything on this sensor, readings and settings alike, is a holding register. That is also what makes `RequestConfigWrite` possible — the sensor's own configuration is reachable through ordinary bus arbitration, as a `COMMAND` that jumps the polling queue.

**Distance is published only while the sensor calls the measurement valid; the state is published on every poll.** A held reading next to a state that is not `0` is legible; a wrong reading is not. `FB_DFROBOT_MQTT_DISCOVERY_DEVICE` is the DFRobot counterpart to the Eastron one, so the sensor gets its own HA device rather than loose entities on the PLC.

| Entity | Category | `device_class` | `state_class` | Unit |
|:--|:--|:--|:--|:--|
| Distance | — | `distance` | — | mm |
| Measurement state | diagnostic | — | — | — |
| Link quality | diagnostic | — | `measurement` | % |

Distance carries no state class on purpose — it is a position, not a quantity worth averaging.

## Commissioning: the bus asks, the device answers

**A device that ships on a baud rate the bus does not use is inaudible, not slow** — and this is fixable from PLC logic alone, with no configuration tool, no USB adapter and no vendor software. The SEN0492 ships at 115200 with no switch on it, which is what made the problem concrete.

**The program that owns the bus names no device.** It asks each registered device one question and acts on the answer:

```
METHOD GetCommissioning : BOOL          (* on RS485Device *)
VAR_INPUT
	BusBaudrate	: UDINT;
	pRequest	: POINTER TO RS485_CommissionRequest;
END_VAR
```

Five of the six implementers answer in one line — `GetCommissioning := FALSE;`, which is what a device configured by switches or by its own display should say. The SEN0492 answers yes and supplies all of it: the address to probe, the register that holds the rate, **the value encoded the way its own register wants it**, and the rates worth trying. Its baud register takes a code rather than a rate, and the code is the index into the block's own table of selectable rates — so one table answers both "which rates can it use" and "what do I write for this one".

`FB_RS485_COMMISSIONER` drives that, and knows nothing about any device:

```mermaid
flowchart LR
  A["ask device:<br/>GetCommissioning"] -->|FALSE| Z["next device"]
  A -->|"TRUE + what to probe,<br/>what to write"| B["retune to the<br/>next setting"]
  B --> C["probe"]
  C -->|silent| B
  C -->|answers at<br/>the bus rate| Y["nothing to write"]
  C -->|answers elsewhere| D["write the value<br/>the device asked for"]
  D --> Y
  Y --> E["RestoreTuning,<br/>publish the result"] --> Z
```

- **It tries the bus rate first**, so an already-commissioned device costs one exchange and the sequence ends there.
- **It holds the bus controller off** while it runs — a device being hunted for at another rate makes the whole pair unusable — and releases it on the first cycle when nothing needs doing.
- **It scans rather than assumes.** The first version wrote the new rate blind at the documented factory speed. The sensor stayed silent afterwards and there was no way to tell whether the write had missed or had landed and moved the sensor somewhere unexpected — a worse position than before it ran. Nothing is written now into a bus whose contents have not been confirmed.

Two supporting changes carry their own weight:

| | |
|:--|:--|
| `RS485Transport` gains `Retune` / `RestoreTuning` / `LastRxCount` | A bus that can be swept has to be re-tunable, and has to be able to say whether *anything* arrived. The sweep no longer repeats the port, parity and buffer size the transport already knows: it changes one field and asks for the rest back. **`SetStopBitsRaw` is gone** — `Retune` subsumes it. |
| `FB_RS485_BUSCONTROLLER` gains `DeviceCount` / `GetDevice` | Commissioning walks the same list the scheduler serves. A second list would drift, and the failure mode is a device that gets polled but was never commissioned. |

`RS485Variables` loses `SEN0492_BUS_BAUD_CODE` — the device encodes its own value now — and `SEN0492_COMMISSION_ON_BOOT` becomes `RS485_COMMISSION_ON_BOOT`, which is what it always was.

:rotating_light: **The reply to a baud-rate write cannot be heard, and nothing may retry on it.** The device changes speed the moment it accepts the write, so its acknowledgement comes back at the new rate and the old one is deaf to it. A failed write is the expected result of a successful one; a retry loop here spins forever on a device that already obeyed.

### Verified on the bench, after the refactor

Full download, all six assertions green:

```
Asked=4  Requested=1  Written=0  Probes=1     every device asked, one answered
LastReport  probes=1 found=9600/255 maxrx=9 at=9600/255 written=FALSE
timerData.PT=TIME#10s                         from the declaration, no override
SEN0492 966 mm state 0   SDM220 227.98 V 4.56 W   SDM_POWER 5.44 W
bus ok=65 noreply=8 crc=2
```

The retained commissioning report is **byte-identical** to what the hardwired sweep published, and the broker diff across the refactor is 475 retained topics to 475, with only `BUS_COUNTERS` moving — that being the counters reset by the download.

## Verified on hardware

The sensor reads:

```
FB_RS485_SEN0492_1/DIST    904 -> 902 -> 902 -> 901   (mm)
FB_RS485_SEN0492_1/STATE   0                          (valid)
```

Distance and measurement state are announced under a DFRobot device — `mf: DFRobot`, `mdl: SEN0492`, `dev_cla: distance`, `unit_of_meas: mm`, both availability topics. Broker diff across the whole change: **5 new retained topics** (four discovery configs and the commissioning report), nothing removed, nothing changed.

**And the rewired bench has two devices answering at once** — the first time this project has had that. Over one 30 s window, all five assertions passed:

| Device | Reading |
|:--|:--|
| SDM220 | 232.37 V, 4.87 W, 50.056 Hz, 403.279 kWh — all three register blocks, no errors |
| SDM_POWER | 4.80 W from the same meter, agreeing with the SDM220's own active power — the cross-check it exists for |
| SEN0492 | reading and announced, 18 mm |
| SDM630 | `offline`, correctly — no such meter is fitted here |

Bus counters over that window: `ok +21` against `noreply +2` and `crc +0..1`. Before the rewiring no-replies were the majority; what is left is the absent SDM630 being asked politely.

### What was actually wrong

**A and B were swapped.** No Modbus setting would ever have fixed it, and the sweep had already ruled out every rate the sensor's own register can select.

The counter that said so was **`LeadNulls`, moving while `Ok` stayed at zero**. A bus with nothing on it returns no bytes at all; bytes arriving that never frame mean the device is powered and talking and something below Modbus is wrong. That is the distinction the diagnostics exist for, and it is what pointed at the pair rather than at a setting.

Once reversed, the sweep found the sensor on its **first** probe:

```
probes=1 found=9600/255 maxrx=9 at=9600/255 written=FALSE
```

- `255` is the sentinel for ordinary framing — so **one stop bit is fine** and the forum reports of 8N2 do not apply here.
- `maxrx=9` is exactly the length of a two-register function 3 reply.
- `written=FALSE` because it was already at 9600: an earlier sweep had moved it, and **that survived several PLC restarts**, so the baud rate is stored in the sensor rather than held in RAM.

:rotating_light: **Whether it survives the sensor's own power cycle is untested** — the bench never cut power to it. If one comes back at 115200, the sweep finds it and moves it again on the next PLC start, which is why it runs every boot rather than once.

## It cannot be polled every two seconds

The connectivity sensor flapped in Home Assistant. The cause was not wiring and not framing: the sensor answers **about once every six or seven seconds however often it is asked** and drops the rest.

| Poll interval | Requests / 30 s | Answered | Success |
|:--|--:|--:|--:|
| 2 s | ~19 | ~4 | **~25%** |
| 7 s | ~4 | ~4 | ~100% |
| 10 s | 3 | 3 | ~100% |
| 30 s | 1 | ~0.7 | ~70% |

`FB_init` now clamps the interval to a measured floor of **8 s**. A caller asking for something the device cannot sustain gets a working sensor and a slower one; the alternative is a reading that is right a quarter of the time.

**The 30 s row is a later measurement, and it narrows this conclusion rather than repeating it.** Slowing down past the floor buys nothing — link quality sat in the same **60–75%** band at 30 s as at 8 s. The rate mattered enormously at 2 s and stops being the limiter at about 8 s, so this project polls at **10 s** and takes the readings rather than the wait.

### The interval needed a second way in

`FB_init`'s `DataPollingInterval` is the honest home for it, but CODESYS stores an instance's `FB_init` arguments outside the declaration text, so no script can revise one — a one-way door for tooling, on the one number most likely to need tuning. Hence two inputs:

| Input | Type | What it does |
|:--|:--|:--|
| `PollIntervalOverride` | TIME | Zero means use `FB_init`'s value; anything else wins, subject to the same 8 s floor on both routes in. Set from `RS485_INIT`, or online while the PLC runs. |
| `AvailabilityFailLimit` | INT | Consecutive failed transactions before `offline`, default 3. Recovery is immediate on the first good transaction regardless. |

:white_check_mark: **The shipped instance now declares `T#10S`, and runs at it.** That correction was made in the IDE, because no script can revise an `FB_init` argument, and it was confirmed where it counts — in the export's `InputAssignments`, not in the declaration text, which is the half that lies. `RS485_INIT` therefore no longer sets `PollIntervalOverride` at all, and `timerData.PT` reads `TIME#10s` on the running PLC with nothing overriding it. The input stays: it is still the only way to retune this without opening the IDE.

**Two wrong answers came first**, both argued from counters rather than bytes, and both are recorded in the commits:

1. A resynchronising `Frame()` — measured identical before and after. It fixes a real class of fault (echoed request, leftover byte, two replies in one window) so it is kept, but it was not this fault.
2. Unregistering the absent Eastron devices — made it *worse*, because the sensor then got polled five times as often. Which is what finally pointed at the rate.

What settled it was capturing the bytes. The transport records the length, address, function and first 24 bytes of any reply that will not frame, and `PLC_PRG_RS485` publishes them with the bus counters to `.../RS485/BUS_COUNTERS`, retained — reading a diagnostic over an online session requires a working online session, which this bench does not reliably have.

## Two signals, because availability cannot do both jobs

Availability is binary and has to be debounced or it reports noise — which means a sensor answering two polls in three reads `online`, true and useless.

| | `/availability` | `/QUALITY` |
|:--|:--|:--|
| answers | can it be reached | how well is it working |
| shape | binary, debounced, on change | percentage, no hysteresis, every poll |

`/QUALITY` is the share of the last sixteen transactions that succeeded, announced as a **diagnostic** entity that still carries `%` and `state_class: measurement` so Home Assistant graphs it. It earned its place immediately: **60–75% while availability reported `online`**.

That required `MQTT_DISCOVERY_SENSOR_DIAGNOSTIC` to gain `unit_of_meas` and `stat_cla`, and `CreateDiagnosticSensorEntity` two optional inputs. **This is the widest-reaching edit here** — every discovery-capable block shares that base. It is additive, the parameters default to empty, and existing diagnostic entities were verified on the broker to pick up the two fields as `null` with nothing else changed.

:rotating_light: **A residual failure rate remains, it is specific to this sensor, and slowing the poll does not help.** Every failed frame has the same shape:

```
received  00 04 00 12 00 00 1A F3
expected  50 03 04 00 12 00 00 crc crc
```

Eight bytes where nine are expected, carrying the sensor's real data — `00 12` is the 18 mm it was reporting at the time — with the first **two** bytes `50 03` collapsed into a single `00`. A byte is genuinely lost on the wire, so no amount of resynchronising in software can recover it: the CRC covers the address, and the address is what got destroyed.

**And there is now a control on the same pair.** The SDM220 runs at roughly **90%** success with `CrcFail` barely moving, so this is not the wiring in general — it is something about how this sensor turns its driver on. Most likely turnaround timing: the SEN0492 answers very fast, so its first byte can begin while the PLC's own driver is still enabled, where a slower meter would not collide. Termination and bias are still the standard remedy and this bus still has neither. Next step is a soldering iron.

## Framing is swept alongside the baud rate

Those are the two things that silence a device on a working pair, so a routine that hunts for one has to try both. Stop bits go in as the numeric SysCom code rather than a `SYS_COM_STOPBITS` enumerator — the enumerator names are not portable across runtimes, and a sweep has to be able to try codes it cannot name. That started life as `SetStopBitsRaw` and is now one field of `Retune`, alongside the rate and the two timeouts. (`Settings` itself is a plain `VAR` and unreachable from outside; the compiler said so and the apply guard refused to save it.)

The result is published **retained** to `.../RS485/SEN0492_COMMISSION`. The obvious place to read a commissioning result is an online debug session, and on this bench that is the least reliable thing in the chain — the runtime drops the session long before it drops MQTT. A result nobody can read is not a result.


docs/FunctionBlocks/FB_RS485_DFROBOT_SEN0492_MQTT.md
```diff
@@ -0,0 +1,332 @@
...
+### **General**
+
+Reads a **DFRobot SEN0492** laser rangefinder over Modbus RTU and publishes the distance,
+optionally announcing itself to Home Assistant. The sensor measures 4 cm to 4 m with roughly
+±2 cm accuracy over RS485.
+
+Distance and measurement state sit in adjacent registers (`16#34` and `16#35`) and are read in
+one request, so the two always describe the same measurement rather than two moments a poll
+apart. Everything on this sensor — readings and settings alike — is a **holding register**, so
+it is function 3 to read and function 6 to write, not the input registers the Eastron meters use.
```
MichielVanwelsenaere: > Everything on this sensor — readings and settings alike — is a **holding register**, so
it is function 3 to read and function 6 to write, not the input registers the Eastron meters use.

to detailed, remove.

docs/FunctionBlocks/FB_RS485_DFROBOT_SEN0492_MQTT.md
```diff
@@ -0,0 +1,332 @@
...
+
+DFRobot documentation:
+- [Product wiki](https://wiki.dfrobot.com/Laser_Ranging_Sensor_RS485_4m_SKU_SEN0492)
+- [Modbus reference and register table](https://wiki.dfrobot.com/sen0492/docs/21034)
+
+----------------------------
+
+:rotating_light: **It ships at 115200 baud and this bus runs at 9600.** There is no switch on
```
MichielVanwelsenaere: > this bus runs at 9600

that's the sample of the bench, the user could use anything. keep it to the essense: we can overwrite it in the PLC logic, no additional tooling needed.

docs/FunctionBlocks/FB_RS485_DFROBOT_SEN0492_MQTT.md
```diff
@@ -0,0 +1,332 @@
...
+inaudible. `PLC_PRG_RS485` scans for it at startup and writes the new baud rate into it; see
+[A device that ships on the wrong baud rate](../RS485/UsingModbusRTU_CODESYS3S.md) for how that
+works and why it scans rather than assuming.
+
+----------------------------
+
+:white_check_mark: **Verified on hardware.** Runs on a CODESYS 3 PFC200 against a real SEN0492:
+distance reads and tracks, `OUTPUT_STATE` reads `0`, and both entities appear in Home Assistant.
+
+:bulb: **If it answers nothing at all, check A and B before anything else.** This sensor was
```
MichielVanwelsenaere: to detailed, remove.

docs/FunctionBlocks/FB_RS485_DFROBOT_SEN0492_MQTT.md
```diff
@@ -0,0 +1,332 @@
...
+
+The declaration is the whole configuration — Modbus address and polling interval through
+`FB_init`, the Home Assistant name through the initialiser:
+
+```
+FB_RS485_SEN0492_1 : FB_RS485_DFROBOT_SEN0492_MQTT(16#50, T#10S)
+                   := (FriendlyName := 'Cistern level');
+```
+
+`16#50` is the factory slave address and collides with nothing else on this bus, so there is no
```
MichielVanwelsenaere: again, that's specific to this project. could not be the case for the user. In addition; 16#50 does not mean anything to a normal user, translate to an int value.

docs/FunctionBlocks/FB_RS485_DFROBOT_SEN0492_MQTT.md
```diff
@@ -0,0 +1,332 @@
...
+probes=1 found=9600/255 maxrx=9 at=9600/255 written=FALSE
+```
+
+Read that as: the sensor answered on the **first** probe, at 9600 with `255` — the sentinel for
+the ordinary framing — so **one stop bit is fine**, despite forum reports of 8N2. `maxrx=9` is
+exactly the length of a two-register function 3 reply. `written=FALSE` because it was already at
+9600: an earlier sweep had moved it, and **that setting survived several PLC restarts**, so the
+baud rate is stored in the sensor rather than held in RAM.
+
+:rotating_light: **Whether it survives the sensor's own power cycle has not been tested** — the
```
MichielVanwelsenaere: this works, remove.

docs/FunctionBlocks/FB_RS485_DFROBOT_SEN0492_MQTT.md
```diff
@@ -0,0 +1,332 @@
...
+where the ceiling actually is: slowing down past the floor buys nothing, because what remains is
+not a rate problem. Asking faster changes nothing except the proportion that fail —
+which is what made the Home Assistant connectivity sensor flap, because three consecutive
+misses is easy to reach at 2 s and hard at 8 s.
+
+A caller who asks for something the device cannot sustain gets a working sensor and a slower
+one. The alternative is a reading that is right a quarter of the time and an entity that keeps
+going unavailable, which is worse in every way that matters.
+
+:bulb: **The shipped instance in `RS485Variables` still declares `T#2S`, and does not run at
```
MichielVanwelsenaere: fix this so it can be removed from the variables and this entire docs section can be removed.

docs/RS485/UsingModbusRTU_CODESYS3S.md
MichielVanwelsenaere: remove this entire section.

docs/FunctionBlocks/FB_RS485_DFROBOT_SEN0492_MQTT.md
MichielVanwelsenaere: Removed. The fact survives once, in the section about changing the sensor's settings, where holding registers are the reason `RequestConfigWrite` works at all.

docs/FunctionBlocks/FB_RS485_DFROBOT_SEN0492_MQTT.md
MichielVanwelsenaere: Rewritten to the essence: it ships at 115200, there is no switch, and the PLC moves it from logic alone - no configuration tool, adapter or vendor software. No bus rate quoted. It also drove the refactor further down this branch: the sweep no longer names this sensor at all, it asks each registered device through `RS485Device.GetCommissioning`, and the bus rate is now one constant in `PLC_PRG_RS485`.

docs/FunctionBlocks/FB_RS485_DFROBOT_SEN0492_MQTT.md
MichielVanwelsenaere: Removed from the block page. It is a wiring diagnosis rather than a fact about the sensor, so it lives once, on the Modbus RTU page.

docs/FunctionBlocks/FB_RS485_DFROBOT_SEN0492_MQTT.md
MichielVanwelsenaere: Example now declares `80`, with `16#50` noted only as DFRobot's own notation. Also dropped "collides with nothing else on this bus" - that is this bench's address map, not the sensor's business; it now says to change it only if something on your bus already answers there.

docs/FunctionBlocks/FB_RS485_DFROBOT_SEN0492_MQTT.md
MichielVanwelsenaere: Removed, and the sentence above it now says positively that the setting survives both PLC restarts and the sensor's own power cycle.

docs/FunctionBlocks/FB_RS485_DFROBOT_SEN0492_MQTT.md
MichielVanwelsenaere: Fixed rather than documented. You corrected the declaration to `T#10S` in the IDE; I confirmed it landed in the export's `InputAssignments` rather than only in the declaration text, removed `PollIntervalOverride := T#10S` from `RS485_INIT`, and deleted the whole caveat from the page. On the running PLC `timerData.PT` reads `TIME#10s` with nothing overriding it. The input itself stays - it is still the only way to retune an `FB_init` argument without the IDE.

docs/RS485/UsingModbusRTU_CODESYS3S.md
MichielVanwelsenaere: Removed. (The API pinned this at line 197 but reported diff position 1, so I checked which section you meant rather than guessing - the e!COCKPIT one is gone, the baud-rate section stays and has been rewritten around the generic mechanism.)

PR #186, opened 2026-08-19
Read a Pt1000 off the RTD module — blocked on configuring the 750-463

MichielVanwelsenaere:
> ## :construction: Draft — blocked on configuring the module, not on the code
>
> **The block, the discovery and both fault paths are finished and verified on hardware. The temperature number is not, and cannot be until the 750-463's channel is reconfigured — which nothing available here can do.**
>
> The bench module is not factory default: channel 0 sits at exactly `1500` — a confident 150.0 °C — and has never moved a digit, while the three unwired channels read `0` rather than over-range. The mapping is right (`%IW1` carries the constant intact); the channel simply is not measuring what is plugged into it.
>
> ### What was tried, and why each is a dead end
>
> | Route | Outcome |
> |:--|:--|
> | **Register communication from IEC code** — write the module's config registers over the process image | **Not on offer.** The CODESYS device description for `01CF_75x_463` (`C:\ProgramData\CODESYS\Devices\288\…\device.xml`) defines exactly one layout: four input words, no control/status byte. There is no mailbox to write through and no IDE setting that adds one. |
> | **WAGO-I/O-CHECK over Ethernet** | **Impossible by design.** Scanning the controller's IP returns *"the communication protocol is not supported by this device or it's deactivated in the device"*, and WAGO's own firmware config explains it: `iocheckd` is an on-demand service **bound to localhost**, fed by the serial service interface (`PTXCONF_IO_CHECK_RS232_STARTLINE="localhost:wago-serv-ser … iocheckd serial"`). Nothing network-facing serves that protocol. |
> | **WAGO's WBM "Ports and Services"** | **Not present.** This controller serves the *CODESYS* WBM; none of WAGO's `config-tools` CGI endpoints answer, so there is no I/O-Check service to enable there. |
> | **WAGO-I/O-CHECK over the service cable** | **Viable, untried** — needs a 750-923 (USB) or 750-920 into the 4-pin header under the front flap, and the PLC application stopped, because the runtime is built with `PTXCONF_CDS3_IODRVKBUS=y` and owns the K-bus. |
>
> ### What would unblock it
>
> 1. The service cable, plus I/O-CHECK pointed at the virtual COM port rather than an IP — then set the channel to **Pt1000, 0.1 °C, 2-conductor, enabled**.
> 2. Or **move the sensor to channel 2, 3 or 4**: if only channel 0 was reconfigured, a neighbour may be untouched, and it is a one-line change here (`Raw := RTD_002`).
> 3. Or sidestep the module entirely — a 1-Wire probe on the ESERA gateway publishes a temperature over the RS485 bus that is already commissioned.
>
> **Open question, being researched separately:** how a 750-463 is configured on a **PFC200 G1** running WAGO firmware with CODESYS Control SL on top — whether the service cable is genuinely the only path, or whether there is a G1-specific route nobody here has found yet.
>
> Merging is not blocked *technically* — the block is additive, compiles clean, and the fault flags behave correctly on the misconfigured channel. It is left as a draft because the headline claim, *this reads a temperature*, is still unproven.

---

The first analog input block here, and the first whose sensor is wired to the PLC rather than reached over a bus. The bench has had a **750-463** in slot four all along — four RTD channels, already mapped as `RTD_001`…`RTD_004` — and nothing in the project ever read them.

```
FB_AI_RTD_001 : FB_INPUT_TEMPERATURE_RTD_MQTT := (FriendlyName := 'Buffer tank top');
FB_AI_RTD_001(Raw := RTD_001);
```

That is the whole configuration. The module puts **tenths of a degree Celsius** in the process image, so there is no scaling factor to get wrong and nothing to calibrate; the block wires itself from `MqttVariables` and announces a temperature sensor plus a diagnostic fault flag to Home Assistant.

Nothing lives in `FB_init`, deliberately — every tunable is an input with a default, so none of it is stuck behind the IDE.

## Two ways a reading can lie, and only one of them looks wrong

**A broken wire is not a temperature.** The module reports a dead sensor by driving the channel to the end of its range, and both which end and which value depend on the module and how it was configured. So the block does not look for a fault code; it asks the question with a stable answer — *could a platinum RTD produce this reading at all?* Outside -200…850 °C it holds the last good value, stops publishing, and raises `/FAULT`. `expire_after` at three heartbeats then lets Home Assistant retire the entity itself.

**And a wrong number can hold perfectly still.** This is the one the first hardware run found, in the block rather than the wiring:

```
RTD_001   1500  1500  1500  1500  1500  1500  1500     seven samples, 70 seconds
```

A confident **150.0 °C**, published with `dev_cla: temperature` and `/FAULT OFF`. 150 °C is a temperature a Pt1000 can reach, so it passed every check the block had. What gives it away is that **a real channel is never still** — the last digit of an RTD reading dithers continuously, and 0.1 °C is far below the noise floor of any room.

Hence `StaleTimeout` (15 minutes by default) and a `Stuck` output. A channel that has not moved a digit in that window raises `Fault` and publishes `/FAULT ON` whatever it is holding — but **keeps publishing the value**, because the number may be right and what is in doubt is whether anything measured it. Silence would claim more than is known.

| | out of range | stuck |
|:--|:--|:--|
| `Valid` | FALSE | TRUE |
| `Fault` → `/FAULT` | ON | ON |
| `/TEMP` | held, not published | still published |

## Verified on the bench

MQTT, end to end — 4 new retained topics, nothing else on the broker touched:

```
Devices/PLC/Lab/Out/AnalogInputs/FB_AI_RTD_001/TEMP    150.0
Devices/PLC/Lab/Out/AnalogInputs/FB_AI_RTD_001/FAULT   ON
homeassistant/sensor/Wago_PFC_Lab_FB_AI_RTD_001/config
homeassistant/binary_sensor/Wago_PFC_Lab_FB_AI_RTD_001_FAULT/config
```

with `dev_cla: temperature`, `unit_of_meas: °C`, `stat_cla: measurement`, `exp_aft: 900`, and the fault flag as a `problem` entity with `ent_cat: diagnostic`. The stale detection was verified against the misconfigured channel with `StaleTimeout` written down to 30 s: `Stuck=TRUE`, `Fault=TRUE`, `/TEMP` still carrying `150.0`.

:rotating_light: **The temperature itself is not verified, and cannot be until the module is reconfigured.** The bench's 750-463 is not factory default: the channel does not track the sensor. Everything measured says so — a value constant to the digit, unwired channels reading `0` rather than over-range, `%IW1` carrying the constant intact. What it needs is WAGO-I/O-CHECK (Pt1000, 0.1 °C, 2-conductor, channel enabled), and there is no path to it from IEC code as this project is configured: the module's control/status byte is not in the process image, so there is no mailbox for register communication.

Two tests that need no tooling are in the guide — move the sensor to another channel, or short the terminals briefly. A channel that is really measuring cannot ignore either.

## Documentation

- `FB_INPUT_TEMPERATURE_RTD_MQTT.md` — the block, both fault modes, and what neither can catch (a wrong sensor type that still tracks).
- `docs/AnalogInputs/UsingRTDSensors.md` — which module, wiring (a Pt1000 has no polarity; the 463 is 2-conductor), who has to configure what, and a commissioning table where **whether the value moves matters as much as what it says**. It also records the measured behaviour that contradicts the data-sheet reading I started from: an unwired channel here reads `0`, not over-range.

## Known wart

The cyclic call sits in `PLC_PRG_MAIN.READ_PUSHBUTTONS`, whose name is now half a lie. An analog channel is a physical input and that action is the one the chart runs every cycle to take inputs in, but it deserves an action of its own — and an action has to be associated with a step in an SFC chart to be called at all, which a script cannot do. The comment in the code says so.


MichielVanwelsenaere:
Superseded by #194.

This branch predates #187 through #193, so the block referenced `MqttVariables`, `FB_MqttPublishQueue`, `InitMqttDone`, `MQTTPublishTopic`, `_InstancePath` and `MqttPublishTopicPrefix` — none of which exist any more. Rebasing would have been a rewrite rather than a conflict resolution, so the block was ported onto current `master` instead and renamed to [`docs/CodingStyle.md`](../blob/master/docs/CodingStyle.md).

Everything here is carried across, including both fault modes and the finding behind `StaleTimeout` — a channel constant to the digit at `1500`. #194 adds the bench run for the renamed instance and its new entity ids, and is honest about one gap this PR implied was covered: the out-of-range branch was never exercised on hardware, and cannot be from a download spec.

The 750-463 blocker is unchanged and documented in `docs/AnalogInputs/UsingRTDSensors.md`.

PR #187, opened 2026-08-20
Give a cover a position, not just three buttons

MichielVanwelsenaere:
A cover Home Assistant can send **to a position**, with a slider instead of three buttons — something this project has not had until now.

```
FB_DO_COVER_002 : FB_OUTPUT_COVER_POSITION_MQTT := (FriendlyName := 'Living room shutter');
FB_DO_COVER_002(T_TravelUp := T#20S, T_TravelDown := T#18S, MU => DO_005, MD => DO_006);
```

## Where the position comes from

There is no encoder on a shutter motor, so the position is **integrated from run time** by `OSCAT_BUILDING.BLIND_CONTROL_S` — a library this project already references, so this needs no new dependency and no forked one. That block keeps a simulated 0..255 position, extends the motor briefly at each end stop so every full journey recalibrates, and holds a lockout across direction changes.

Reading its documentation — which this repo also already vendors — was worth the detour: **`UP` and `DN` both TRUE is not a contradiction, it is the block's automatic mode**, where it regulates its position to the setpoint on its own. That is the whole trick this block is built on. What is left here is translation: percent to and from 0..255 with rounding, Home Assistant's payloads and topics, and deciding what a stop means.

## Three interruptions, deliberately not the same

| | effect |
|:--|:--|
| `STOP` over MQTT | motors off, and **the setpoint is dragged to the present position** — a stop that silently resumes later is not a stop |
| releasing `UP` / `DN` | the same hold, so a person watching the cover always wins over what Home Assistant asked for |
| `PRIO_LOCK` | motors off, setpoint **kept**: whatever was asked for happens when the lock clears |

## The discovery layer learns about position

`MQTT_DISCOVERY_COVER` gains `pos_t`, `set_pos_t`, `pos_open` and `pos_clsd`; `CreateCoverEntity` takes them as optional parameters. **Supplying `set_position_topic` is what puts a slider on the entity**, so it is the one field that matters on the Home Assistant side. Empty publishes `null` for all four — the same convention `dev_cla` and the units already follow.

`MqttSubCoverTopic` ends in `#` rather than `+` now: the position command topic sits one level below the cover's own topic, which a single-level wildcard would silently never deliver.

## Verified on hardware, end to end

Coils on a 750-540 with LEDs on the outputs, commanded from the broker and from the online interface. All fifteen assertions green:

| | |
|:--|:--|
| at rest after a download | both coils off, `PositionKnown` FALSE, `STOPPED` — **a restart moves nothing** |
| `OPEN` | up coil on, `OPENING`, position 20 → 51 |
| `STOP` mid-travel | both coils off, position held at 54 |
| `CLOSE` | down coil on, `CLOSING`, 34 |
| reaching the bottom | 0, **`PositionKnown` TRUE**, `CLOSED` |
| set position 60 | up coil on, arrives at 58, stops itself inside `Tolerance` |
| throughout | `DO_003`, `DO_004`, `DO_007`, `DO_008` untouched |

From the broker, a set-position of 35 while resting at 58:

```
58  CLOSING  53  48  43  38  STOPPED
```

**The Home Assistant fix is confirmed where it matters:** `FB_DO_COVER_001`'s config now carries no position keys **at all** — absent, not null — so HA stops rejecting it, while `FB_DO_COVER_002` carries `pos_t`, `set_pos_t`, `pos_open: 100`, `pos_clsd: 0`. Every other cover on that broker, including the house and the stable, is untouched. 482 retained topics before and after.

### Three bugs the LEDs caught, and the compiler could not

Two of them would have passed for a working cover on a dashboard:

1. **The lockout deadlocked.** The timer that grants permission to start was fed the decision it was meant to gate, so asking to move reset it and the request was refused forever. Symptom: `STOPPED` at 0% whatever was asked.
2. **It drove itself open on every restart** — the target defaulted to 100 and nothing recorded that no request was outstanding. The first bench run opened the cover to 42% before being asked anything.
3. **A full open stopped at 98%**, because the tolerance was applied to end-stop targets too: the motor never reached the stop, the estimate never recalibrated, and the entity read `STOPPED` where it should have read `OPEN`.

### Where the coils came from

`750-540` is a **4**-channel module, so the bench has eight digital outputs and all eight were already claimed — `DO_005`…`DO_008` by `PLC_PRG_HVAC`, from `HvacTask`. This branch had been writing two of them from `MainTask`, i.e. two writers on one coil in two tasks, with no compiler diagnostic whatsoever. `HVAC_RUN` now assigns no physical output (its logic and MQTT entities untouched), cover 001 keeps `DO_003`/`DO_004`, and this block gets `DO_005`/`DO_006`. Every coil has exactly one writer in one task.

## Two deliberate departures from the older cover block

- **It drives `BLIND_CONTROL_S` directly, not through OSCAT's `BLIND_INPUT` stage.** Pushbutton handling here belongs to `FB_INPUT_PUSHBUTTON_MQTT`, and using OSCAT's input stage as well would put two state machines in charge of one shutter.
- **Travel times are inputs, not `FB_init` arguments.** It is the one number on a cover that always gets tuned after watching it move, and an `FB_init` argument cannot be revised by tooling.


README.md
```diff
@@ -40,6 +40,7 @@ Basic function blocks for basic IO events and operations.
 - [FB_OUTPUT_BINARY_MQTT](./docs/FunctionBlocks/FB_OUTPUT_BINARY_MQTT.md)
 - [FB_OUTPUT_DIMMER_MQTT](./docs/FunctionBlocks/FB_OUTPUT_DIMMER_MQTT.md)
 - [FB_OUTPUT_COVER_MQTT](./docs/FunctionBlocks/FB_OUTPUT_COVER_MQTT.md)
+- [FB_OUTPUT_COVER_POSITION_MQTT](./docs/FunctionBlocks/FB_OUTPUT_COVER_POSITION_MQTT.md) - a cover Home Assistant can send to a position, with a slider rather than three buttons
```
MichielVanwelsenaere: > - a cover Home Assistant can send to a position, with a slider rather than three buttons

remove, don't mention any text next to the link, just as the other links.

docs/FunctionBlocks/FB_OUTPUT_COVER_POSITION_MQTT.md
```diff
@@ -0,0 +1,247 @@
...
+| publish | `.../Out/Covers/<instance>/POSITION` | `0`..`100`, every `PublishStep` while moving and exactly once movement ends |
+| subscribe | `.../In/Covers/<instance>` | `OPEN`, `CLOSE`, `STOP` |
+| subscribe | `.../In/Covers/<instance>/POSITION` | `0`..`100` |
+
+:bulb: **The cover subscription is a `#` wildcard because of this block.** The position command
+topic sits one level below the cover's own topic, and `MqttSubCoverTopic` used to end in `+`,
+which would not have delivered it. The older cover block's topics still match, so nothing changed
+for it.
+
+:white_check_mark: **Verified on hardware, end to end.** A CODESYS 3 PFC200 with the coils on a
```
MichielVanwelsenaere: remove R149 to R186

README.md
MichielVanwelsenaere: Removed - the entry reads like every other one in that list now.

docs/FunctionBlocks/FB_OUTPUT_COVER_POSITION_MQTT.md
MichielVanwelsenaere: Removed, lines 149-186: the bench narrative and the list of bugs found on the way. It was a record of an afternoon rather than documentation of a function block, same as the provenance note earlier.

The test those bugs came from only existed in gitignored .ai/ scratch, so it would have died with the session. It is now `specs/cover-position.json` in the test-plc-logic skill, with a section naming the three traps that each produced a plausible-looking cover - a restart must move nothing, a full open must reach 100 and say OPEN, and STOP must not be a pause. Re-run from its new path: nine steps green, plus the broker half (58 CLOSING 53 48 43 38 STOPPED).

PR #188, opened 2026-08-21
Bring the DALI block back into the reference project

### LINKED ISSUE #178: Evaluate DALI on G1 and bring the block back into the reference project
## Where DALI stands today

DALI is half-present in the reference project, and the half that is missing is the half that gets tested.

- `FB_OUTPUT_DIMMER_DALI_MQTT` is **not in the project**. It lives only as a standalone export in `src/Exports/archive/FB_OUTPUT_DIMMER_DALI_MQTT.xml`.
- `DALIVariables` **is** still in the project — as an empty husk. Its entire content is one commented-out line:
  ```
  {attribute 'qualified_only'}
  VAR_GLOBAL
  	// M1_Light1: typBallast:=(bAddress:=0,xIsGroup:=FALSE,bPortDALI:=1);
  END_VAR
  ```
- `docs/FunctionBlocks/FB_OUTPUT_DIMMER_DALI_MQTT.md` exists and the README lists it as `(archived)`.
- Nothing references `FbDaliMaster` anywhere in the project.

## First: is "requires G2" actually true?

The README says DALI *"requires a G2 PFC device, and is therefore not part of the reference project"*. **That should be evaluated before anything else is built around it** — it is reported to work on a **G1** device under e!COCKPIT, and a library is a library.

Evidence that supports re-testing rather than assuming:

- The **G1 PFC200 K-bus device description** already contains the module. In `C:\ProgramData\CODESYS\Devices\288\0000 0001\4.19.0.0\device.xml` — the same catalogue the reference project's `Pfc200Bus` is built from — there is an entry `0287_75x_647`, described as *"75x-647 DALI"* / *"DALI Multi-Master Module"*, alongside `75x-641 DALI/DSI`. So the module can be placed on a **G1** bus in CODESYS 3.5.
- The working e!COCKPIT project resolves `FbDaliMaster` from the **`WagoAppDALI`** library (`WagoAppDALI, * (WAGO)`).
- The reference project's Library Manager has **no** `WagoAppDALI` reference, and no DALI library is vendored under `src/Libraries/` (MQTT, OSCAT, PRO_JSON, SysFile23/SysSocket23 only).
- `WagoAppDALI` is **not installed** in this CODESYS 3.5 installation — nothing matching `*Dali*` exists under `C:\ProgramData\CODESYS`.

So the blocker looks like the **library**, not the controller generation. If that is right, the fix is obtaining `WagoAppDALI` for CODESYS 3.5 rather than buying different hardware.

### How to settle it, no hardware needed

1. Obtain and install `WagoAppDALI` for CODESYS 3.5 (WAGO package / library repository), and vendor it under `src/Libraries/` like the other dependencies.
2. Add the library to the reference project's Library Manager.
3. Add a `753-647` to `Pfc200Bus` in the device tree.
4. Import `src/Exports/archive/FB_OUTPUT_DIMMER_DALI_MQTT.xml` and run `./tools/ai/codesys.ps1 verify`.

That answers the compile half completely. Only ballast addressing, groups and fade behaviour need a real module afterwards.

## Why it is worth fixing regardless of the answer

The archived block is **never compiled**. It cannot appear in `verify`, it cannot appear in a baseline, and a refactor that touches `FB_MQTT_BASE`, the discovery structs or `CreateSensorEntity` — all of which the DALI block uses — will not fail against it. It will simply drift until someone imports it into a real project and finds out.

That has already nearly happened. The recent `stat_cla`/`StateClass` work (#177) changed `MQTT_DISCOVERY_SENSOR` and `CreateSensorEntity`; the archived block was outside every check performed. It happens to still be fine, but nothing verified that.

Worth knowing: the archived export was diffed against the copy running in a real installation and the bodies are **identical**. The block is current and working — it is the reference project that lost it, not the code that rotted.

## What done looks like

1. **Evaluate the G1 route above.** If it compiles, DALI comes back into the reference project directly and this issue is mostly finished.
2. **If it genuinely needs G2**, fall back to adding a G2 device node to the reference project that carries the block. A CODESYS project can hold several devices; a second controller node that is never downloaded still gets compiled, which restores the missing coverage without hardware.
3. **Decide what `DALIVariables` is for.** Either populate it, or remove it — an empty GVL with one commented line reads like support exists.
4. **Correct the README** once the answer is known. Either the G2 claim goes, or it gains the real reason.
5. Drop `(archived)` from the README and the function block page, and let `update-fb-docs --check` cover that page again.

## Open question

- Is a G1 + 753-647 combination available on the bench to confirm the runtime half, once the compile half passes?

MichielVanwelsenaere:
Closes #178.

`FB_OUTPUT_DIMMER_DALI_MQTT` lived only as a standalone export under `src/Exports/archive/`, so it was never compiled — it could not appear in a `verify`, could not appear in a baseline, and a refactor of `FB_MQTT_BASE` or the discovery structs would not have failed against it. #178 asked whether *"requires G2"* was actually true.

**It was not the blocker. The library was.**

## What was wrong

`WagoAppDALI` simply was not installed. It ships in WAGO's **Device Support Package**, whose version is pinned to the CODESYS patch level — `2.0.8.9` for our `3.5.21.30`. With it installed, the block compiles on a **G1** device under *CODESYS Control for PFC200 SL*, and the `753-647` turns out to be in that runtime's own K-bus catalogue already.

One code change was needed: the library is **qualified-only**, unlike e!COCKPIT, so the four DALI type references are now `WagoAppDALI.typBallast`, `WagoAppDALI.FbDaliSendDimValue` and so on.

## Why there is now a second controller

Landing the block in the pool is *not* the same as compiling it — a POU nothing instantiates is skipped, which is the exact trap this issue was about. So the project gains **`Wago_PFC200_G2_Virtual`**: a 750-8212 with a `753-647` on its K-bus, never downloaded, carrying `PRG_DALI_VERIFY`. That program instantiates the block and binds `FbDaliMaster.I_Port` to the real module — a binding a G1 device cannot make, because only WAGO's own device description implements `I_Port_753_647`. An application that is never downloaded is still compiled, so the whole chain is checked on every build.

The lab device is renamed `Wago_PFC200_G1_Lab` and is otherwise untouched; its K-bus is unchanged.

**The ninth baseline warning is load-bearing.** `PRG_DALI_VERIFY.Dimmer.DimValue` having no persistent list is the only evidence in a plain build that the DALI block is genuinely compiled. If it ever disappears, the verification has stopped verifying — it is documented as such rather than silenced.

## Two caveats, now written down

- **Only a G2 WAGO device can drive the module under the CODESYS runtime.** The module can sit on a G1 bus, but the SL description gives 24 raw bytes and no driver function block, so there is nothing for `I_Port` to bind to. The interface is a stateful mailbox abstraction, not a byte port, so re-implementing it over those raw bytes was considered and rejected.
- **DALI is untested on hardware.** No ballast addressed, no group configured, no fade timing measured. Compile-verified only.

## Not vendored, deliberately

WAGO's licence forbids redistributing the library ("publicly reproduce or make the Software available", "pass on the Software to third parties") and this repo is public — so `docs/WagoPfcPrep.md` carries a one-off per-machine install instead, with the package/CODESYS version pairing, and the toolchain tables list it as a dependency of the build itself.

## Tooling

All `-Force`-gated and all refusing to save a project that does not build:

- `device -AddDevice type:id:version [-Under node] -NodeName x` — a module, or a whole second controller
- `device -RemoveNode` / `device -RenameNode`
- `scaffold -Scaffold spec.json` — creates GVLs, programs and tasks inside an application, since adding a device yields an application with a Library Manager and nothing else. `tools/ai/scaffold/g2-dali-verify.json` rebuilds this one.

Two harness bugs surfaced on the way, both invisible while the project had a single application and both making it report **less than it had checked**:

| Bug | Effect | Fix |
|:--|:--|:--|
| `build()` is incremental | an up-to-date application compiled nothing, said nothing, still landed in `built` | `rebuild()` |
| a build **clears** the message store | sweeping once at the end returned only the *last* application's messages — a baseline briefly came back with none of the eight known warnings | collect per application |

The second fix also un-swallowed the MQTT library's `Information` notes, which are in the baseline now.

## Verification

- `verify -Baseline` → `RESULT: OK`, both applications built, 9 warnings (8 known + the load-bearing one)
- `verify` → `NEW vs baseline: 0`
- `update-fb-docs --check` → clean; the block's page is generated again instead of frozen
- One step needed the IDE and cannot be scripted: enabling dynamic memory on the new application. Recorded in `CLAUDE.md` with the three property names that were tried and rejected.


docs/FunctionBlocks/FB_OUTPUT_DIMMER_DALI_MQTT.md
```diff
@@ -9,45 +11,82 @@ DALI configuration via the Wago DALI tool and creation of the `typBallast` is ex
...
+:rotating_light: **Needs the `WagoAppDALI` library, which is not vendored in this
+repository.** WAGO's software licence forbids redistributing it, so it has to be
+installed once per engineering PC from WAGO's own package — see
+[Installing the WAGO libraries](../WagoPfcPrep.md#installing-the-wago-libraries-dali).
+Without it the project does not build at all.
 
-It is kept as a standalone PLCopen export so it can be imported into a G2 project: **[src/Exports/archive/FB_OUTPUT_DIMMER_DALI_MQTT.xml](../../src/Exports/archive/FB_OUTPUT_DIMMER_DALI_MQTT.xml)**. See [Restoring it into a project](#restoring-it-into-a-project) below. It is unverified on CODESYS — please report back if you get it running.
+The library is **qualified-only**, so its types are referenced as
```
MichielVanwelsenaere: remove this, not relevant to reader

docs/FunctionBlocks/FB_OUTPUT_DIMMER_DALI_MQTT.md
```diff
@@ -9,45 +11,82 @@ DALI configuration via the Wago DALI tool and creation of the `typBallast` is ex
...
+`WagoAppDALI.typBallast`, `WagoAppDALI.FbDaliSendDimValue` and so on. Code copied
+from e!COCKPIT, where the bare names resolve, will not compile here until it is
+qualified.
+
+`DimValue` is persistent, so a new instance also needs its instance path in
```
MichielVanwelsenaere: same here remove these three lines.

docs/FunctionBlocks/FB_OUTPUT_DIMMER_DALI_MQTT.md
MichielVanwelsenaere: Removed in e9a49d9. Agreed — that was maintainer trivia on a page meant for someone using the block. The qualified-only naming moved to the `codesys-loop` skill, next to the WagoAppDALI dependency row, since that is where it matters when ST gets lifted out of an e!COCKPIT project. The `DimValue` persistent note is already covered by that skill's baseline section.

docs/FunctionBlocks/FB_OUTPUT_DIMMER_DALI_MQTT.md
MichielVanwelsenaere: Removed in e9a49d9. Agreed — that was maintainer trivia on a page meant for someone using the block. The qualified-only naming moved to the `codesys-loop` skill, next to the WagoAppDALI dependency row, since that is where it matters when ST gets lifted out of an e!COCKPIT project. The `DimValue` persistent note is already covered by that skill's baseline section.

Issue #179, opened 2026-08-16
Introduce a naming convention for every datatype, and write it down


MichielVanwelsenaere:
## The problem

The project has grown three or four naming styles for the same kind of thing, and there is nothing written down that says which one is right. Every new block re-opens the question, and every answer makes the inconsistency slightly worse.

This is what is in `src/Exports/PLCopen.xml` today.

**Enumerations** — two conventions, neither dominant:

| Name | Style |
|:--|:--|
| `E_MQTT_ENTITY`, `E_RELAY_TYPE` | `E_` prefix, SCREAMING_SNAKE |
| `HvacModes` | PascalCase, plural, no prefix |
| `RS485_EASTRON_SDM_Devices` | domain prefix + mixed case, plural |
| `DIMMER_CURVE` | SCREAMING_SNAKE, no prefix, singular |
| `Dimmer` | PascalCase, no prefix |

**Structures** — mostly SCREAMING_SNAKE, with exceptions:

- `MQTT_DISCOVERY_*` (22 of them), `MQTT_MESSAGE` — consistent
- `RS485_RtuQuery` — domain prefix, then PascalCase

**Function blocks** — the `FB_` prefix is universal, the rest is not:

- `FB_OUTPUT_BINARY_MQTT`, `FB_RS485_EASTRON_SDM630_MQTT`, `FB_MQTT_LOG` — SCREAMING_SNAKE
- `FB_MqttPublishQueue`, `FB_MqttPublishWorker` — PascalCase

**Interfaces** — one exists, `RS485Device`, with no marker at all. Nothing tells a reader it is an interface rather than a function block.

**Global variable lists** — `MqttVariables`, `DMXVariables`, `DALIVariables`, `RS485Variables` … and `PersistentVars`.

**Variables** — a partial Hungarian prefix scheme, applied where it was copied in and absent where it was not: `pMqttPublishQueue`, `xComPortOpen`, `awReadBuffer`, `stComPortSettings`, `hComPort`, `eComError`, `bStripJsonRootOk` sit next to `Trigger`, `InitRS485`, `Startup` and `ActiveRtuQuery`, which are the same datatypes without the prefix.

## Scope: the whole project, no grandfathering

This lands as part of the **major release**, so the convention applies to everything. No "new code only" carve-out — a standard that half the project ignores is not a standard, and the major version is the one chance to pay for it in a single, announced break.

## What done looks like

[...]

PR #191, opened 2026-08-23
Stop the pump running on against a shut manifold

MichielVanwelsenaere:
On a collector, the pump could keep turning after every valve had been commanded shut. Whether it actually got there depended on two timings configured on two different blocks, with nothing in the code relating them — and the signal needed to fix it already existed and was read by nobody.

## The fault

`FB_HVAC_PUMP_MQTT` wraps `OSCAT_BUILDING.ACTUATOR_PUMP`, whose switch-off branch is gated on the minimum on-time:

```
ELSIF pump AND NOT in AND NOT manual AND tx - last_change >= min_ontime THEN
```

So the pump does **not** follow `IN`: once running it keeps running until `MIN_ONTIME` has elapsed. That is the point of the block — a pump that short-cycles wears out. But `FB_HVAC_COLLECTOR_MQTT` assigned `VALVE[i] := THERMOSTAT[i]` unconditionally, so it closed the whole manifold in the cycle demand ended, and the pump spent the rest of its minimum on-time pushing against valves that were closing.

The condition, exactly:

> dead-head if the valves close in less than the pump's `MIN_ONTIME`

which is to say it depended on `ValveCycleTime` on one block against `MIN_ONTIME` on another, chosen at two separate declaration sites, with no assertion, no comment and no compile-time relationship between them. The reference sits on the safe side by a minute — `T#3M` against `T#2M` — but nothing defended the margin. Worse, `ValveCycleTime` describes a valve **opening**; how long one takes to *shut* is not modelled anywhere, so the margin rested on a physical property the code never sees.

The collector itself was never inconsistent: `PUMP` implies `bHeatRequest` implies at least one valve open. The gap was only ever at the composition boundary in `PRG_HVAC`, which is why nothing inside either block looked wrong.

## It was a known requirement, and half-built

`PRG_HVAC.HVAC_RUN` has carried the specification since the chain was designed. 2.2 was implemented; 2.3 was not:

```
// 2. The Thermostat connects to a heating Valve turning it on/open when heat is required
//    2.1 The heating valve takes X amount of time to switch fully open
//    2.2 The heating valve can only turn on the pump once it's fully open
//    2.3 The heating valve can only start closing once the pump has completed its minimum cycle time
```

And `FB_HVAC_PUMP_MQTT` already computed the signal for it:

```pascal
MIN_ONTIME_ACTIVE := NOT(fbMinOntimeTimer.Q) AND PUMP;
```

TRUE exactly during the run-on, because the TON resets when `IN` falls while `PUMP` is still held. Written every cycle, read by nothing in the project.

## The fix

One input on the collector, wired to the output that already existed:

```pascal
fbPump2Collector.PUMP_MIN_ONTIME_ACTIVE := fbPump2.MIN_ONTIME_ACTIVE;
```

While it is set, the circuits that were last flowing are held open, so the pump always has a path. `abValveHeld` remembers which ones, so the water keeps going where it was already going rather than through whichever circuit happens to be index 1.

**Backwards compatible by construction.** The input defaults to FALSE, which is the old behaviour precisely, so an existing call site that does not wire it keeps compiling and keeps closing its valves the moment demand ends.

**The load-bearing comment is now a warning, and it is the review-worthy part of this PR.** `bHeatRequest` is built from `THERMOSTAT` and deliberately **not** from `VALVE`:

```pascal
IF bHeatRequest THEN
    FOR iI := 1 TO ciValveCount DO
        VALVE[iI]      := THERMOSTAT[iI];
        abValveHeld[iI] := THERMOSTAT[iI];
    END_FOR
ELSE
    FOR iI := 1 TO ciValveCount DO
        VALVE[iI] := PUMP_MIN_ONTIME_ACTIVE AND abValveHeld[iI];
    END_FOR
END_IF
```

Those two ORs used to be the same value — every `VALVE[iI]` was `THERMOSTAT[iI]`, and a comment in the block said so. They are not the same value any more. Rebuild `bHeatRequest` from `VALVE[]` and the held valves re-request the pump, which is then held on for ever. The old comment inviting that simplification is replaced by one forbidding it.

Which is also why the hold terminates: it re-opens valves, but the block never re-requests the pump, so the minimum on-time expires, the input clears, the valves close.

## The test is a regression test, and one number makes it one

`specs/hvac-valve-pump-interlock.json` sets **`MIN_ONTIME` (40 s) longer than `ValveCycleTime` (10 s)**. That is the configuration in which the fault appears. At the production values the pump stops before the valves finish closing anyway, so the same nine steps pass on broken code — a test at production timings here proves nothing, and that is stated at the top of the spec so nobody "corrects" it later.

Step 5 is the assertion. Step 7 asserts the hold **releases**, because a latch is the failure the fix itself could introduce, and proving it is better than reasoning about it.

**9/9 on the lab PFC200:**

```
[ok] 5. THE REGRESSION.
     fbThermostat2.OUT=FALSE        bHeatRequest=FALSE
     fbPump2.PUMP=TRUE              MIN_ONTIME_ACTIVE=TRUE
     fbPump2Collector.PUMP=FALSE    PUMP_MIN_ONTIME_ACTIVE=TRUE
     fbPump2Collector.VALVE[1]=TRUE

[ok] 7. ... fbPump2.PUMP=FALSE  VALVE[1]=FALSE  VALVE[2]=FALSE
```

No demand, nothing requesting the pump, pump still turning, valve still open — then both clear, valve after pump. Steps 8 and 9 restart the chain normally, so the ordinary path is intact.

Broker at **480 retained topics before and after, nothing in `NEW` or `GONE`** — no discovery config orphaned, no entity created or lost. `verify -Baseline` unchanged at 0 errors and the committed 9 warnings.

The honest limit: a bench has no manifold, so this asserts *commanded* valve state and measures no real valve travel.

## Three things the hardware run taught, all now in the skill

It took three attempts and neither failure was the code. Each fails in a way that looks like a broken PLC, which is why they are written down rather than remembered:

- **An enum is written as its ordinal and read back fully qualified**, so one variable needs two spellings in the same step. `"E_HVAC_MODE.off"` raised `'E_HVAC_MODE#E_HVAC_MODE.off' is not a valid Integer` on five steps at once.
- **`delay_ms` is a floor, not the elapsed time.** Every `read` and `expect` is a round trip to the PLC, so a step with ten assertions adds about a second of its own. Assertions placed 2 s from a 5 s timer edge drifted past it and failed twice against a working interlock. Every edge now has 7 s of margin.
- **Force the state you are about to drive, not just the state around it.** `eHvacMode` and `rDesiredTemp` are PERSISTENT and survive a download's cold reset, so `fbThermostat2` arrived already at `MODE heat` with a 20.5 setpoint and started the chain a step before the spec meant to. Setting the *other* thermostats off is not enough.

## Also in here

**The pump page documents the run-on**, which nothing did before: that `HEAT_REQUEST` is an alias of `PUMP` and therefore carries the run-on into the burner, and that `RUN_EVERY := T#0S` is deliberate — OSCAT's anti-seize feature would start the pump with every valve shut, since no thermostat asked for anything and the collector cannot know. That last one is a latent trap rather than a bug today, and it is one line away from being real.

**The HVAC guide gets the timing rule**, because the three timings live at three declaration sites and two of them interact. Scenario A is not exposed — the thermostat drives the pump with no valve in between. B and C are.

**`test-plc-logic`'s HVAC prose named pre-#179 instances** — 52 occurrences, so every topic and variable in it was wrong, and the names were fixed from the project's own declarations rather than guessed. Its step 7 also described the old valve behaviour as the correct outcome.

One doc/code disagreement corrected on the way past: the pump page claimed `HEAT_REQUEST` "follows the input `IN` more closely than output `PUMP`". The code says `HEAT_REQUEST := PUMP`, and the block's own comment says that alias is deliberate.

## Not in this PR

**The burner is the same fault one level up.** It has its own independent minimum on-time, so its window can start later than the pump's and therefore end later — the boiler still firing after the pump has stopped. Chaining two minimum on-times does not compose into one. Whether it matters depends on the boiler's own pump overrun, which is outside what this project can see, so it is left as a known item rather than guessed at.

**Closing time is still not modelled.** `ValveCycleTime` is the opening delay and the docs still describe it as "open or close". With the interlock wired the two timings no longer have to be chosen relative to each other, so this is now a documentation wart rather than a hidden dependency — but it is not fixed.

## For anyone with a collector in their own project

After syncing these blocks you get the old behaviour until you add one line, because the input defaults FALSE. If you leave it unwired, keep the pump's `MIN_ONTIME` shorter than the time your valves take to close, and know that nothing checks it for you. On real hardware the robust answer is both: wire the interlock **and** fit a differential bypass on the manifold, which gives the pump a path regardless of what the control logic believes.


docs/FAQ/HVAC_getting_started_guide.md
```diff
@@ -30,4 +30,40 @@ Not all buildings leverage floor heating in every single room. In many setups ra
...
+| `MIN_ONTIME` / `MIN_OFFTIME` | `FB_HVAC_PUMP_MQTT` `FB_init` | Minimum run and minimum pause, so the pump cannot short-cycle. |
+| `MIN_ONTIME` / `MIN_OFFTIME` | `FB_HVAC_BURNER_MQTT` `FB_init` | The same, for the heat source. |
+
+**The pump's minimum on-time outlives the demand that started it.** Once the pump
+is running it keeps running until `MIN_ONTIME` has elapsed, however quickly the
+thermostat is satisfied. So on a collector, whatever closes the valves has to wait
+for the pump rather than for the thermostat — otherwise the pump spends the rest of
+its minimum on-time pushing against a manifold that is closing.
+
+The fix is to wire it, not to tune it:
```
MichielVanwelsenaere: not fix, approach

docs/FAQ/HVAC_getting_started_guide.md
```diff
@@ -30,4 +30,40 @@ Not all buildings leverage floor heating in every single room. In many setups ra
...
+
+```ST
+fbPump2Collector.PUMP_MIN_ONTIME_ACTIVE := fbPump2.MIN_ONTIME_ACTIVE;
+```
+
+That is what requirement 2.3 in `PRG_HVAC.HVAC_RUN` asks for, and with it in place
```
MichielVanwelsenaere: remove this entire section, not relevant.

docs/FAQ/HVAC_getting_started_guide.md
MichielVanwelsenaere: Changed to "The approach is to wire it, not to tune it". Agreed — this page describes how the blocks are meant to compose, so calling it a fix both dates the text to one PR and reads a design decision as a defect.

docs/FAQ/HVAC_getting_started_guide.md
MichielVanwelsenaere: Removed. It was PR rationale — citing requirement 2.3 in HVAC_RUN and arguing about an unmodelled closing time — not something a reader getting started needs.

I kept the two paragraphs after it (the Scenario A/B/C exposure note and the differential-bypass tip), on the reading that your other remark asks for a wording fix at line 53 and so the section itself stays. Say the word if you meant the whole "Getting the timings right" section and I will drop it.

Nothing is lost from the docs: the practical half — leave the input unwired and MIN_ONTIME has to be shorter than the valves closing time — is already on FB_HVAC_COLLECTOR_MQTT's page, where someone deciding whether to wire it is actually looking.

PR #192, opened 2026-08-23
Make user LED 1 report broker health, instead of nothing

MichielVanwelsenaere:
U1 was meant to be green while the PLC was talking to the MQTT broker and blinking red while it was not. Reading the code, it had never done either — and the mechanism that was supposed to make it work turned out not to be enough even once it did.

## Two faults, neither visible to the compiler

**The green branch had no input.** `PRG_MQTT` declared `stMQTTInfo : MQTT.MQTT_INFO` and drove its edge detectors off `stMQTTInfo.MQTT_CONNECTED` — but nothing ever populated the struct. `MQTT_INFO` is an **output** of the library's `MqttClient`, not something that fills itself in, so the flag was `FALSE` from the first cycle of the first commit that had it. `git log -S` says it was never assigned in the whole history.

**The red branch was never called.** `fbMQTTDisconnectTrigger` was an `F_TRIG` whose `.Q` was read every cycle in `MQTT_PUBLISH` and which was *invoked* only from `MQTT_INIT` — an `IF NOT initMqtt` block that runs once. An edge detector called once cannot produce an edge.

Both compile perfectly, both look right on the page, and the documentation described the feature as working.

## What the library actually exposes, and how that was established

The MQTT library is a compiled `.library` — the objects inside are CODESYS's own binary serialisation, so `strings`, UTF-16 scans and zlib probing all come back empty. It was settled by **compile probe** instead: 13 candidate member paths in one throwaway `verify`, reading which lines produced no error.

```
[Error] 'MQTT_INFO' is no component of 'MQTT_IN_OUT'          <- line 46
                                                              <- line 49, silent: valid
[Error] 'MQTT_CONNECTED' is no component of 'MqttClient'      <- line 55
                                                              <- line 56, silent: valid
```

Two survivors: `MQTT_IN_OUT.clientFB.MQTT_INFO` and `MQTT_IN_OUT.clientFB.CONNECTED`. A second probe that tried to *write* them both got `'MQTT_INFO' is no input of 'MqttClient'`, which is how you learn something is a `VAR_OUTPUT` when you cannot read the source. `MQTT_INFO` carries `MQTT_CONNECTED : BOOL` and `MQTT_ERROR : STRING`.

## The fix, and why it is not two edge detectors any more

```pascal
stMQTTInfo := MQTT_IN_OUT.clientFB.MQTT_INFO;

bBrokerHealthy := stMQTTInfo.MQTT_CONNECTED AND GVL_MQTT.bBrokerReachable;

IF bBrokerHealthy <> bLedShowsHealthy THEN
    IF bBrokerHealthy THEN
        PFC.SetLed(which := PFC.LED.U1, how := PFC.LedState.STATIC_GRN);
    ELSE
        PFC.SetLed(which := PFC.LED.U1, how := PFC.LedState.BLINK_RED);
    END_IF
    bLedShowsHealthy := bBrokerHealthy;
END_IF
```

`SetLed` is still called only on a change, but from a remembered state rather than a pair of detectors. **One code path cannot have half of it left uncalled, which is precisely how this broke.** And `bLedShowsHealthy` reports what the LED is showing, which a one-cycle `Q` cannot — so a test can read the colour back.

`MQTT_INIT` sets U1 red and `bLedShowsHealthy := FALSE` once, because a change detector writes nothing until something changes: without it a PLC that never reaches the broker shows whatever the runtime left on the light.

## The part a cable found

With the above landed and the bench test green, Michiel pulled the LAN cable and **U1 stayed green.** That is correct behaviour from the library and wrong behaviour for the LED: TCP has nothing to report when a cable comes out — no FIN, no RST — so the socket stays open until a retransmit finally gives up, minutes later. The client goes on claiming a session the entire time.

So `MQTT_CONNECTED` cannot be the whole condition. `PRG_PING_DMX` now pings the broker too and writes `GVL_MQTT.bBrokerReachable`; U1 is green only while the client has a session **and** the host answers. The two measurements cover each other's blind spot:

| | ping | `MQTT_CONNECTED` | U1 |
|:--|:--|:--|:--|
| all well | answers | TRUE | green |
| cable out, switch dead, broker host down | fails | stays TRUE for minutes | **red, from the ping** |
| host up, broker process stopped | answers | goes FALSE (clean TCP close) | **red, from the client** |

The broker host comes from `GVL_MQTT.broker` with the port stripped, rather than a second constant that could drift from the one the client dials.

## `SysSockPing` returns a third thing, and it matters

The first version tested `= 0`. A test run then read `udiBrokerReachable = 1` while the broker was answering and the flag said reachable — and the pre-existing U3 code, it turns out, has always handled only `0` and `5` and quietly ignored everything else. A later run pointed at an unreachable address and returned **`5`**, confirming `5` is the real failure code and `1` is noise.

So the verdict is debounced: two consecutive failures before the broker counts as unreachable, one success to clear it.

```pascal
IF udiBrokerReachable = 0 THEN
    uiBrokerPingFails := 0;
ELSIF uiBrokerPingFails < cuiBrokerPingFailsForRed THEN
    uiBrokerPingFails := uiBrokerPingFails + 1;
END_IF
GVL_MQTT.bBrokerReachable := (uiBrokerPingFails < cuiBrokerPingFailsForRed);
```

Counting beats enumerating return codes: an unknown code still counts as a failure, but no single one of them can flip a cabinet light. **Without this, that stray `1` would have blinked U1 red for no reason.** Red lands within ~20 s of a cable coming out, green within ~10 s of it going back in; the Ping task period is the knob if that is too slow.

## Testing something that has no readable state

`PFC.SetLed` is a runtime call. No variable holds a colour, so the only witness for the light is a person at the cabinet — and that half is confirmed by eye: red on unplug, steady green on replug.

Everything that *decides* the colour is assertable, and `specs/mqtt-broker-led.json` drives it through a real failure rather than a forced flag. Pointing `GVL_MQTT.broker` at **`192.0.2.1`** — reserved by RFC 5737, so it can never be a host — makes the ping genuinely fail while leaving the MQTT session up. That is the pulled-cable case exactly, and nothing in the path is stubbed: ICMP, the debounce, the flag, the AND, the change detector and the real `SetLed` all run.

**3/3 on the lab PFC200 (`00E8`):**

```
[ok] 2. THE RED BRANCH.
     PRG_PING_DMX.sBrokerHost='192.0.2.1'   udiBrokerReachable=UDINT#5
     PRG_PING_DMX.uiBrokerPingFails=UINT#2  GVL_MQTT.bBrokerReachable=FALSE
     PRG_MQTT.bBrokerHealthy=FALSE          bLedShowsHealthy=FALSE
     PRG_MQTT.stMQTTInfo.MQTT_CONNECTED=TRUE      <- session still believed up
```

That last line is the whole test: connected in the client's opinion, unreachable in fact, LED red. Step 3 then heals on its own — nothing writes the flag back, so recovery is the Ping task doing it, which makes step 3 the test that the ping runs at all.

Broker at **480 retained topics before and after, nothing in `NEW` or `GONE`.** A light and a cover commanded over MQTT to prove the subscription path still works end to end. `verify -Baseline` unchanged at 0 errors and the committed 9 warnings.

## Two dead ends, recorded so nobody repeats them

- **Writing `GVL_MQTT.bBrokerReachable` FALSE from the harness.** Racy, and it failed exactly that way once: the Ping task owns that flag and rewrites it every 10 s, so the assertion chases another task's value. Drive the input, not the output.
- **Pointing `broker` at a dead *port* to break the session.** The write lands and reads back, and 25 s later `MQTT_CONNECTED` is still `TRUE` — the client does not act on a URL change while its socket is up. Which is annoying for testing and *useful* for the real test, because it is what keeps the session up in step 2.

## Also in here

- **A reconnect now republishes availability immediately.** The session's rising edge really fires, so `online` no longer waits up to 5 s for the heartbeat. Same root cause, same one-line fix.
- **`MQTTClient` is gone** — a whole unused `MQTT.MQTTClient` instance declared in `PRG_MQTT` and referenced nowhere, which made "where does the connection state live?" harder to answer than it needed to be. So is the commented-out `//LED_1_BrokerStatus: FbAppLED;` in `PRG_PING_DMX`, someone's earlier run at this feature.
- **The docs described the feature as working**, with pre-#190 variable names. Rewritten, with the three ways to end up with a clean build and a dark LED.
- **`test-plc-logic` claimed `availability` must read `online`.** It cannot: the birth message is published `MqttRetain := FALSE` while the LWT is retained, so a retained snapshot of that topic reads `offline` whatever the PLC is doing. Corrected to `-Watch`, and noted in the MQTT docs.

## Not in this PR

`MQTT_INFO.MQTT_ERROR` is a string the library fills in and nothing reads. It is the obvious input for a diagnostic log message and it is left alone.

Detection is bounded by the 10 s Ping task, so ~20 s for a cable pull. Fine for a cabinet indicator; if it ever needs to be faster the task period is the thing to change, and that affects the U3 ping too, which is why it is not changed here.


docs/AdditionalFunctionality/MQTT_General.md
```diff
@@ -117,3 +118,20 @@ Topic: `Devices/PLC/Lab/availability`</br>
 Payload: `offline`
 
 Note that the topics and payloads can be changed in the code. The Birth message is by default published during startup and after that every 5 seconds (heartbeat). This can be changed in the code as well.
+
+The Birth message is also published on connect and on **re**connect, off the rising
+edge of the client's `MQTT_CONNECTED` flag in `PRG_MQTT` - the same edge that drives
+the [U1 LED](User_leds_CODESYS3S_runtime.md). Until that flag was actually populated
```
MichielVanwelsenaere: > Until that flag was actually populated
the edge could never fire and the heartbeat was doing all the work, so a reconnect
went unannounced for up to five seconds.

remove, not relevant

docs/AdditionalFunctionality/MQTT_General.md
```diff
@@ -117,3 +118,20 @@ Topic: `Devices/PLC/Lab/availability`</br>
...
+
+:bulb: **A retained snapshot of `availability` reads `offline` even while the PLC is
+online.** The Birth message is published with `MqttRetain := FALSE` while the LWT is
+retained, so the last retained value on that topic is whichever `offline` the broker
+published when the client last dropped. To find out whether the PLC is really
+connected, watch the topic live instead of reading the retained value:
+
+```powershell
+./tools/ai/Mqtt-Snapshot.ps1 -Watch -Seconds 12 -Topics 'Devices/PLC/Lab/availability'
+```
```
MichielVanwelsenaere: > 
:bulb: **A retained snapshot of `availability` reads `offline` even while the PLC is
online.** The Birth message is published with `MqttRetain := FALSE` while the LWT is
retained, so the last retained value on that topic is whichever `offline` the broker
published when the client last dropped. To find out whether the PLC is really
connected, watch the topic live instead of reading the retained value:

```powershell
./tools/ai/Mqtt-Snapshot.ps1 -Watch -Seconds 12 -Topics 'Devices/PLC/Lab/availability'
```

remove, not relevant

docs/AdditionalFunctionality/MQTT_General.md
MichielVanwelsenaere: Removed. Kept the first sentence — that the Birth message is published on connect and reconnect off that edge — since it describes what the page is actually about. The part I cut was explaining why it did not used to work, which belongs in the PR, not on a reference page.

docs/AdditionalFunctionality/MQTT_General.md
MichielVanwelsenaere: Removed, whole block. The fact itself still matters — a retained read of that topic is the LWT, not a status — so it stays in the test-plc-logic skill under "Before you start", which is the thing that sends someone to look at availability in the first place. It was wrong there before this PR (it said the topic must read online), so that correction is worth keeping; it just does not belong on this page.

PR #193, opened 2026-08-24
Give each device its own Log and Availability, and stop truncating names

MichielVanwelsenaere:
## Why

Every discovery device got an `Availability` and a `Log` entity for free from
`FB_BASE_MQTT_DISCOVERY_DEVICE.initBaseDevice` — and both named the **controller's**
topics. With six discovery devices in this project that is six `Log` sensors
subscribed to one shared stream and six `Availability` sensors all mirroring the
PLC's LWT. Five of each said nothing about the device they hang off.

It looked like "every function block gets a Log sensor". It isn't — ordinary entity
blocks create nothing diagnostic. The count scales with *discovery devices*, and
four of the six are owned by RS485 device blocks.

## What changed

**Diagnostics follow the device.**

```
Devices/PLC/Lab/diagnostic/Log                              <- the PLC (unchanged)
Devices/PLC/Lab/Out/RS485/<instance>/diagnostic/Log         <- each device
Devices/PLC/Lab/Out/RS485/<instance>/availability           <- each device
```

The four self-owning RS485 blocks pass `CONCAT(sMQTTPublishTopic, '/diagnostic')`,
mirroring the PLC's own `<base>diagnostic` shape. An RS485 device's `Availability`
entity now reports what `OnTransactionDone` publishes — whether that device answers
on the bus — instead of whether the controller is up.

The 1-Wire sensor's discovery device lives in `GVL_MQTT` and is configured through
`FB_init`, whose arguments a script cannot reach. So `init1WireDevice` gained two
optional inputs and the sensor block passes its own topics at the call site;
`FB_init` is untouched and no `InputAssignments` were put at risk.

**`FB_MQTT_LOG` is deleted.** A pool POU with no instance anywhere, therefore never
compiled, whose doc page described a `logger` member of the discovery device that
has not existed for some time. Its orphaned `collector_FB_MQTT_LOG` in `GVL_MQTT`
goes with it. Logging is `SendLogMessage` on the discovery device.

**Widths**, because IEC string assignment and `CONCAT` both truncate in silence:

| | |
|:--|:--|
| 93 declarations | every stored complete topic → `MQTT_TOPIC_LEN` |
| 22 declarations | every instance-path holder → `MQTT_TOPIC_LEN` |
| 1 declaration | `SendLogMessage`'s `instance`, which cut the widened path back to 80 |

The instance-path one is the load-bearing find. The longest in this project was
**73 characters of an 80-character default** — seven to spare — and those feed
`FB_MQTT_BASE._sInstancePath`, from which `DeviceName` and every publish topic
suffix are derived. CODESYS truncates an instance path from the **left**, so an
overflow puts the first dot inside a `..` marker and `DeviceName` becomes garbage,
with a clean build and no warning.

Also fixes `'as MQTT dicovery'` in the log payload.

## Verification

Compile-checked (`NEW vs baseline: 0`, 9 pre-existing warnings) **and run on the
bench**, since a clean build says nothing about which topics appear:

- 5 new per-device `Log` topics, each carrying a real payload — so `SendLogMessage`
  reaches them, not just compiles
- all 10 diagnostic configs repointed; **no `GONE`**, so nothing orphaned — entity
  ids are unchanged
- the 1-Wire sensor's own availability propagated to all four of its entities
- instance paths publish in full, at 94 characters
- the deletion of `FB_MQTT_LOG` building clean is itself the proof nothing used it

The `SendLogMessage.instance` truncation was found **on hardware after the first fix
looked right in source** — the `..` had simply moved to the other end of the string.

## One number worth having

A discovery payload is composed by PRO_JSON, whose value field is
`STRING(GPL_JSON.MAX_VALUE_SIZE)`, and **`MAX_VALUE_SIZE` is 150**. That is the real
ceiling above our own declarations. Not reachable here (~70 characters), but it is
the number to check before inventing long device names. It cannot be read from the
library, so it was established by compile probe; the method and its traps are in
`MQTT_General.md`, including that the obvious probe — an over-long string constant —
reports the field as too small for a 21-character literal and is simply wrong.

## Known, not addressed

- The diagnostic pair inherits the two-topic availability list, so `Availability`
  reads `connected` or goes `unavailable` — it can never show `disconnected`, and a
  device's `Log` is unreadable exactly when the device is down. Narrowing `avty` for
  those two entities needs an override on `FillCommonFields`.
- The `Name` / `Id` / `Identifiers` family is still at the `STRING(80)` default.
  Entity ids currently run to 65 characters, so **15 to spare**; a truncated
  `uniq_id` would give duplicate Home Assistant entities. Same class of fix, its own
  change.
- `POINTER TO STRING` topic prefixes are left alone — a deref reads to the NUL, so it
  is a different failure mode, and widening them means touching pointer types at
  every `ADR()` site.
- `FB_RS485_DUCO_DUCOBOX_MQTT` has no `InitMqttDiscovery` at all, so it announces no
  device and gets no diagnostics.
- The 1-Wire discovery device is shared from `GVL_MQTT`; with one sensor that is
  fine, a second would share the first's identity.

Installation projects need a re-sync to pick the blocks up.


docs/AdditionalFunctionality/MQTT_SelfWiring.md
```diff
@@ -184,6 +184,7 @@ call, the prologue has nothing to wait for. Configuration through `FB_init` and
 self-wiring through `FriendlyName` are complementary: together they remove the
 whole init sequence rather than half of it.
 
-`FB_MQTT_LOG` extends the base and could self-wire, but it has no call site
-anywhere in the reference project and there is no log topic prefix in
-`GVL_MQTT`, so there was no existing behaviour to preserve. It was left alone.
+`FB_MQTT_LOG` used to be listed here as a block that could self-wire but had no
```
MichielVanwelsenaere: remove no history

docs/AdditionalFunctionality/MQTT_General.md
```diff
@@ -20,43 +20,76 @@ VAR_GLOBAL
...
+
+**The real ceiling is lower than 160, and it is not ours.** A discovery payload is
+composed by `PRO_JSON`, whose `JSONVAR` value field is
+`STRING(GPL_JSON.MAX_VALUE_SIZE)` - and `MAX_VALUE_SIZE` is **150**. A topic longer
+than 150 characters is cut when it is written into the discovery struct, whatever
+our own declarations say. 150 is not reachable in practice here (the longest topic
+in this project is about 66 characters), but it is the number to check against
+before inventing very long device or instance names.
+
+That value cannot be read from the library - a `.library` does not unpack, and
```
MichielVanwelsenaere: remove this entire secion, no value to reader

docs/AdditionalFunctionality/MQTT_SelfWiring.md
MichielVanwelsenaere: Removed. That paragraph is gone rather than rewritten - the block does not exist any more, so there is nothing for a reader to look up. Also removed the same kind of note I had left at the end of the `Device diagnostics` section in `MQTT_General.md`, for consistency.

docs/AdditionalFunctionality/MQTT_General.md
MichielVanwelsenaere: Removed the whole `### Topic lengths` section. Fair - the generated GVL region directly above it already shows `MQTT_TOPIC_LEN := 160`, `MQTT_SUFFIX_LEN := 64` and every declaration using them, so the prose was restating the code.

One thing that leaves nowhere in the repo: `GPL_JSON.MAX_VALUE_SIZE` is **150**, which caps a discovery topic below our own 160 regardless of these declarations. Not reachable today (longest topic ~70), and it is in the PR description, so it is on the record if it ever matters.

PR #194, opened 2026-08-25
Read a Pt1000 off the RTD module, and publish it

### LINKED ISSUE #171: Support 4-channel analog input : 750-460
Hello,
Would it be possible to add support for the 750-460 card to measure temperatures with PT1000 sensors ? 
I use this system with my home automation system and would like to migrate to your project combined with Home Assistant. 
It should be possible to adjust the offset and accuracy (number of decimal places).

MichielVanwelsenaere:
Closes #171.

## What #171 asked for, and what this gives it

The request was to read Pt1000 sensors on a **750-460** and get them into Home
Assistant. The block is module-agnostic — it reads one channel word in tenths of a
degree — so the 750-460 is covered by the same code as the bench's 750-451, which is
what the requester expected: *"in terms of programming, this should not make no
difference (750-460/000-003 for pt1000 and 750-460 for pt100)"*. The installed device
description agrees: `75x-460 4AI, RTD`, four `INT` channels, with `/000-003` as the
Pt1000 variant. It is now named in both module lists, which it wrongly was not.

Two smaller asks in that issue deserve a straight answer rather than a tick:

- **"adjust the offset"** — `LeadResistance` is the offset that matters on a
  2-conductor module, and the 750-460 is 2-conductor. It is given in ohms of lead
  resistance rather than in °C, because that is the quantity you can actually measure
  at the terminals. For a sensor in every room on long cable — the requester's case —
  that is the correction worth having.
- **"accuracy (number of decimal places)"** — **not** configurable, deliberately. The
  module delivers one decimal and no more, so the block formats from the integer
  rather than through `REAL_TO_STRING`, which would publish `21.299999` for a channel
  reading `213`. Offering more decimal places would be inventing precision the
  hardware does not have. If that is a blocker for anyone, it is a separate
  conversation about a different module.

:rotating_light: **Verified on a 750-451, not on a 750-460.** Nobody here has that
module. The claim rests on it being the same family reading the same way, which the
device description and the requester both support — but it is an inference, not a
measurement, and it should be said out loud before the issue is closed.

Supersedes #186, which was written before the naming convention landed in #190
and had drifted too far to rebase — its branch predates #187 through #193, so the
block referred to `MqttVariables`, `FB_MqttPublishQueue`, `InitMqttDone` and
`_InstancePath`, none of which exist any more. This is the same block, ported onto
current `master` and renamed to [`docs/CodingStyle.md`](docs/CodingStyle.md).

## What it does

The bench carries a **750-451** — eight RTD channels, mapped as `RTD_001`…`RTD_008`
— and nothing in the project ever read an analog input at all.

```
fbAiRtd001 : FB_INPUT_TEMPERATURE_RTD_MQTT := (FriendlyName := 'Lab PLC temperature 1');
fbAiRtd001(Raw := RTD_001);
```

That is the whole configuration. The module puts **tenths of a degree Celsius** in
the process image, so there is no scaling factor to get wrong and nothing to
calibrate; the block wires itself from `GVL_MQTT` and announces a temperature
sensor to Home Assistant, which goes unavailable whenever the reading cannot be
trusted.

Nothing lives in `FB_init`, deliberately — every tunable is an input with a default,
so none of it is stuck behind the IDE.

## `LeadResistance`, for a module that cannot compensate

Both RTD modules measure **2-conductor**, so the leads sit in series with the element
and the channel reads high by them. A Pt1000 moves ~3.9 Ω/°C, which sizes the problem:

| Cable | Loop | Error |
|:--|:--|:--|
| 2×10 m of 0.5 mm² | ≈ 0.7 Ω | ≈ 0.18 °C — ignorable |
| 2×25 m of 0.5 mm² | ≈ 1.7 Ω | ≈ 0.44 °C |
| 2×50 m of 0.25 mm² | ≈ 6.8 Ω | ≈ **1.8 °C** — not ignorable |

`LeadResistance` is the measured loop resistance in ohms; the block takes
`LeadResistance / 3.9` degrees off, in tenths and in integers so the module's single
decimal reaches the payload without a REAL round trip. Zero by default, and
`REAL_TO_INT(0.0)` is 0, so an unset channel is bit-for-bit what it was.

An **input with a default, not an `FB_init` argument** — `FB_init` arguments live in
`InputAssignments` where a scripted edit never reaches them, and a value that varies
per installation is the worst possible thing to bury there.

`3.9`, not `3.85`: 0.00385 is the *mean* coefficient over 0–100 °C, while the local
slope at 0 °C is 3.9083 Ω/°C. A lead correction applies where the sensor is, not
averaged across its span.

The movement check deliberately keeps reading the **uncorrected** `Raw`: the
correction is a constant and cannot change whether a channel is moving, and editing
`LeadResistance` on a running PLC must not let the resulting jump look like life.

:rotating_light: **Correct it in one place only.** The module has its own *user
scaling* and *user calibration*, written with WAGO-I/O-CHECK into its flash, which the
PLC cannot read. Set an offset in both and it is applied twice with nothing to say so
— a 1.8 °C fix becomes a 1.8 °C error the other way, reading as a badly calibrated
sensor. Prefer the PLC: the value is in source control, appears in a diff and in the
export, and survives swapping the terminal, which a number in module flash does not.

Verified on the bench, with the arithmetic chosen so the expected answer needs none
— 3.9 Ω is exactly one degree:

```
LeadResistance   Raw   Temperature   /TEMP
0                262   26.2          26.2
3.9              262   25.2          25.2     one degree, as promised
39               262   16.2          16.2     ten - it scales
0                262   26.2          26.2     and it fully releases
```

`Raw` never moves, so the module's word is untouched and only the derived value
shifts.

## A broken wire is not a temperature

The module reports a dead sensor by driving the channel to an end of its range, and both which end
and which value depend on the module: a 750-463 parks at `1500` (150.0 °C), a 750-451 at `8500`
(850.0 °C). **Both are legal platinum temperatures**, so the IEC 60751 range check passes them
— which is why the block also asks the question that has a useful answer: *could this sensor be
measuring this?* Outside `PlausibleMin`…`PlausibleMax` (-40…80 °C by default) it holds the last
good value, stops publishing, and publishes `availability: offline`, so the entity greys out where
somebody is already looking at it.

On the bench that is seven unwired channels reporting `offline` on the first scan, about 34 seconds
after a restart.

### The first reading after a restart was a zero nobody measured

Caught on the broker during review, on a cold start, inside one second:

```
availability offline  →  TEMP 0.0  →  availability online  →  TEMP 26.3
```

Home Assistant recorded the `0.0` as a real reading. A mapped channel does not read
its sensor the moment the application starts — the process image is cleared to zero,
and `0` is 0.0 °C, a perfectly plausible temperature, so the block believed it.

**Two gates now, because there are two different reasons the word is zero.**

| Gate | Answers | Source |
|:--|:--|:--|
| `ProcessImageValid` (new input) | is the bus exchanging process data at all? | `Pfc200Bus.xConfigFinished` |
| `ctFirstConversion` (5 s constant) | has *this module* had time to convert? | timer started when the bus comes up |

The first alone did **not** fix it, and the log said why: `xConfigFinished` reports the
**bus**, not the module, and goes TRUE slightly before an RTD terminal has produced its
first conversion. The fix still published a `0.0`, just later in the same second. Hence
the second gate: a zero is not believed until the module has had time to convert. Any
non-zero reading is trusted at once; only zero waits, and after the timeout zero is
believed like any other value — 0.0 °C is a real temperature and a sensor sitting at it
must not be unavailable forever.

`ProcessImageValid` defaults to **FALSE** deliberately. Forget to wire it and the channel
stays unavailable, which someone notices; defaulting TRUE hands out a wrong number once
per restart, silently — which is the bug being fixed.

This also retires the `DataAvailable` wart documented since the block was written: it
latched on that same pre-K-bus zero, and now cannot.

**Verified on hardware.** Watch armed before the download, so the first publish is
captured:

```
15:17:14  all eight channels   availability -> offline
15:17:15  fbAiRtd001/TEMP      26.3          <- first value published is the real one
15:17:15  fbAiRtd001/availability -> online
15:18:14  fbAiRtd001/TEMP      26.3          (heartbeat)
```

Live `0.0` publishes on any `/TEMP` topic: **0**. Channels 2-8 stay `offline` throughout,
which is correct — they are open circuits. The entity is briefly unavailable after a
restart instead of briefly wrong.

### `StaleTimeout` was removed

The block used to carry a second check: a channel that had not moved a digit for `StaleTimeout`
was flagged `Stuck` and faulted. It is gone, along with the `Stuck` output, and `Fault` is now
simply `NOT Valid`.

It does not survive contact with a quiet room. Measured on this bench, a Pt1000 moved from `259` to
`262` over about three and a half hours — roughly **one digit every 75 minutes**, holding each value
dead still in between. A fifteen-minute timeout sits inside that quiet period around 80% of the
time, and a false fault publishes `availability: offline`, taking a **working** sensor out of Home
Assistant. That is a worse outcome than the failure it was watching for.

And it had already been superseded: the only case it ever demonstrably caught — an open circuit
parked at an end of scale — is rejected on the first scan by `PlausibleMin`/`PlausibleMax`, which
landed later.

The docs now name what is *not* guarded rather than implying the surface is complete: a plausible
frozen value, and a wrong sensor type that still tracks. Both are commissioning problems, answered
with a thermometer and by shorting the terminals once.

Verified on the bench after a cold start — the last line being the point:

```
wired channel      Raw=262   26.2 C   Valid=TRUE   Fault=FALSE   '26.2'
unwired channel    Raw=8500  850.0 C  Valid=FALSE  Fault=TRUE
45 s later, same   Raw=262   26.2 C   Valid=TRUE   Fault=FALSE
```

## What the port changed

Behaviour is unchanged — this is a rename, not a redesign. Pins and method
parameters are the published interface and are exempt by the convention, so
`Raw`, `Temperature`, `Valid`, `Fault`, `Stuck`, `DataAvailable` and the three
tuning inputs read exactly as they did in #186.

| | #186 | here |
|:--|:--|:--|
| private locals | `Published`, `Startup`, `Payload`, `StaleTimer`, `LastRaw`, … | `rPublished`, `bStartup`, `sPayload`, `fbStaleTimer`, `iLastRaw`, … |
| constants | `RTD_MIN_RAW`, `RTD_MAX_RAW` | `ciRtdMinRaw`, `ciRtdMaxRaw` |
| the reflection local | `InstancePath` | `sInstancePath` |
| base members (#190) | `InitMqttDone`, `MQTTPublishTopic`, `_InstancePath`, `MqttPublishTopicPrefix` | `bInitMqttDone`, `sMQTTPublishTopic`, `_sInstancePath`, `psMqttPublishTopicPrefix` |
| the GVL and the queue (#190) | `MqttVariables`, `FB_MqttPublishQueue` | `GVL_MQTT`, `FB_MQTT_PUBLISH_QUEUE` |
| the instance | `FB_AI_RTD_001` | `fbAiRtd001` |
| `MqttPubAnalogInputPrefix` | `STRING(100)` | `STRING(GVL_MQTT.MQTT_TOPIC_LEN)`, per #193 |

The instance rename **changes the Home Assistant entity id**, exactly as #190 did
for every other block. #186's two retained state topics were still on the broker
under the old name with nothing pointing at them; they are cleared, and the
snapshot diff below shows them going.

## Verification

Compile-checked at every step — 0 errors against the 9-warning baseline, `NEW vs
baseline: 0` — and the block is really compiled, because `fbAiRtd001` is declared
and called rather than auto-instantiated by the verify harness.

Then downloaded to bench node `00E8` and read back online:

```
RTD_001       259    25.9 °C   Valid=TRUE  Fault=FALSE  Stuck=FALSE  DataAvailable=TRUE
RTD_002..008  8500  850.0 °C   the seven unwired channels, all identical
```

and, watching the state topic for a further 140 s, `fbAiRtd001/TEMP` published
`26.0` three times — once a minute, which is the heartbeat, and **one digit above
where it started**.

The broker diff is clean: `fbAiRtd001/availability` `offline` → `online`, `TEMP`
`0.0` → `26.0`, no unexpected `GONE`. The only other retained topics to change were
`RS485/BUS_COUNTERS` and `SEN0492_COMMISSION`, which any download resets.

## What the hardware change settled

The three 750-463s this PR was written around were faulty and are off the rail. One
**750-451** — eight channels, 2-conductor — replaced them, so the twelve instances
became eight and `RTD_001..008` are its channels 0..7 in order. That is not a
tidy-up: it is the first reading on this branch that is a measurement rather than an
artefact, and it settles three things at once.

- **The channel is measuring.** The 750-463s held `1500` to the digit for as long as
  anyone watched. This one moved a digit within minutes, which is the signature the
  `Stuck` check was written around.
- **The module is set for Pt1000**, established by measurement rather than a data
  sheet: a Pt1000 on a Pt100 channel saturates, so a correct absolute reading is
  itself the proof of the sensor type. No WAGO-I/O-CHECK session was needed after
  all — which is the cheapest possible outcome of `docs/WagoIoCheck.md`.
- **An unwired channel on a 750-451 saturates high**, at `8500`, where the 750-463's
  disabled channels read `0`. Two modules, two open-circuit values, and 850.0 °C sits
  inside IEC 60751 exactly as 150.0 °C did — so **neither is caught by the range
  check, and both are caught by `PlausibleMin`/`PlausibleMax`.** The seven empty
  channels report `offline` and show as unavailable rather than as seven confident
  850 °C sensors. The out-of-range branch that #186 could not exercise is now
  exercised, seven times over, by real open circuits.

:rotating_light: **Mapping a terminal takes three passes, not one**, and the obvious
order does not work: `device -MapIo` builds before saving and refuses a red build, so
a `PRG_MAIN` referring to channels that do not exist blocks the very mapping that
would create them. Take the references out, map, put them back. Written down on the
RTD page.

## Still outstanding

**Warming the element by hand and watching the value follow.** A dithering last digit
is strong evidence of a live conversion; a number that tracks a deliberate change is
proof. Said rather than implied.

**`exp_aft` on the broker reads `3`, and the code computes `180`.** Chased to a
root cause, and it is worse than it looked — see below.

## Why `exp_aft` never updated: the completeness guard rejects every config

**Since 67d8329, this PLC has not published a single Home Assistant discovery
config.** Not one — not the RTD channels, not the covers, pushbuttons, dimmers,
thermostats, pumps or the burner. Everything Home Assistant shows for
`Wago_PFC200_G1_Lab` is a retained config from before that commit, which is exactly
why `exp_aft` still reads `3`: it was published in the era when the sweep interval
was `T#1S`, and nothing has overwritten it since.

Watched live across a cold start, the two config topics received **nothing**, while
the log said:

```
10:09:39  ERROR Wago_PFC200_G1_Lab_fbAiRtd001 composed truncated json
10:09:39  ERROR Wago_PFC200_G1_Lab_fbDiPb001  composed truncated json
10:09:40  ERROR Wago_PFC200_G1_Lab_fbThermostat1 composed truncated json
          ... every entity on the PLC, plus plc-lab_diag_log and _diag_availability
```

**The JSON is not truncated. The check that says so is wrong.**

```
bJsonComplete := RIGHT(sMqttJSON, 1) = '}';
```

`sMqttJSON` is `STRING(1500)`, and the string functions are declared over
**`STRING(255)`** — so the argument is cut to its first 255 bytes on the way in, and
`RIGHT(…, 1)` returns **character 255**, not the last character. For `fbAiRtd001`
that character is `i`, from `…availability"}],"` in the middle of the availability
array. It is never `}`, so the guard fails for every config longer than 255
characters — which is all of them, on every block.

Note what this is *not*: `FB_MQTT_BASE.bTopicTruncated` guards a different
truncation — a composed topic outgrowing its 160-character buffer. Here the buffer
is not the narrow part at all. `sMqttJSON` has 1500 bytes and uses 665 of them; it
is the **function signature** that is 255 wide, so the value is cut on the way into
a check whose whole job is to detect cut values.

What rules out the alternatives, in the order they were eliminated:

| Theory | Killed by |
|:--|:--|
| The publish queue overflowed during the startup burst | `GVL_MQTT.fbMqttPublishQueue.DroppedCount` read online: **0** at 45 s, 75 s and 105 s |
| Discovery never ran after the cold start | `fbAiRtd001.bInitMqttDiscoveryDone` = **TRUE** |
| `HeartbeatInterval` really is one second | read online: **`TIME#1m`**, and TEMP republishes once a minute |
| `bInitMqttDiscoveryDone` survives a download | `GVL_PERSISTENT` lists specific members only; the flag is not among them |
| The config is too long for the buffer | 665 characters into a `STRING(1500)`; other blocks publish 795 |
| `STRUCT_TO_JSON` really did truncate | `F_STRIP_JSON_ROOT` returned TRUE, and it works on a **pointer** across the whole 1501-byte buffer — it found `:{` and a closing `}` and moved the content. A truncated compose fails it, and it did not fail |

:rotating_light: **67d8329 verified this check against the wrong thing.** Its commit
message reads *"Verified that all 434 retained configs end in exactly `}` with no
trailing whitespace, so the check cannot misfire on real output."* That verification
was run in Python against the broker's retained payloads — which confirmed the
**data** was fine and never exercised the **ST function's** semantics. The check
misfires on precisely the input it was verified against, and it does so silently
except for a log line nobody was watching.

### Fixed, and tested against the bug

The guard now walks to the null terminator through a pointer, the way
`F_STRIP_JSON_ROOT` already does two lines above it. It keeps doing its job — a
genuinely truncated compose ends mid-object, not on a brace.

The compiler cannot catch this class of fault: narrowing to `STRING(255)` is a legal
implicit conversion. So **`tools/ai/check_string_widths.py`** does. It reads the
committed PLCopen export — which carries both every string's declared width and the
ST that uses it — resolves symbolic widths from the GVL constants, and fails when
`LEN`, `LEFT`, `RIGHT`, `MID`, `CONCAT`, `FIND`, `INSERT`, `DELETE` or `REPLACE` is
handed a string wider than 255. Per-method scope plus the POU's interface, so a name
cannot collide across POUs. No CODESYS and no PLC needed.

**It was written against the bug, not after it.** On the pre-fix export it reports
exactly the one real line and nothing else:

```
FB_BASE_MQTT_DISCOVERY_DEVICE.PublishEntityConfig line 27
    bJsonComplete := RIGHT(sMqttJSON, 1) = '}';
    RIGHT() is STRING(255); sMqttJSON is STRING(1500) - the argument is cut to its first 255 bytes
```

On the fixed export it passes, with 19 wide strings examined. `--list` shows what it
watches — `PRG_RS485.sDiagReport` at `STRING(400)` and the publish queue's `Payload`
among them, so there is more than one place this could happen again.

It is deliberately **not** wired into CI: `pull-request.yml` is left alone, and the
check is run by hand alongside the other `tools/ai` scripts.

### And a runtime test, because the compile check would not have caught the damage

The narrowing check finds the one bad line. It would not have found the
*consequence*, because the consequence was not in the code — it was a broker full of
configs that were valid, complete and months old. Nothing here could tell a config
the PLC had just written from one it had failed to write since spring.

**`tools/ai/check_mqtt_discovery.py`** answers both halves.

*Structure*: every config is valid JSON, ends in a closing brace, has a `uniq_id`
that is unique across the broker **and** matches its own topic's node segment, a
`dev` with ids, and something to talk to; `avty` entries carry a topic, `avty_mode`
is one of all/any/latest, `exp_aft` is a non-negative integer. All 430 retained
configs on the bench broker pass.

*Recency*: MQTT has no publish timestamp, so "did this download write it?" looks
unanswerable — but the **retain flag** answers it. Subscribe live: the broker's
replay arrives with `retain=1`, a genuine publish with `retain=0`. Record the
expected set, watch across a download, and anything that never arrives live is a
finding. It also fails on `composed truncated json` / `failed json root strip` /
`had empty MqttJSON` appearing live on the diagnostic log topic — the PLC's own
report of a config it declined to publish, which nothing else reads.

`exp_aft: 0` is **not** a failure — it is Home Assistant's "no expiry" and 195
configs use it deliberately. My first draft failed all of them, which is the wrong
kind of test. A small *positive* expiry is a NOTE instead, and on the bench that note
names exactly the eight stale RTD configs at `exp_aft=3`, with no false failure
anywhere.

:rotating_light: **One bug found in the test by running it, and it is the failure
mode a test must never have.** `%r` is mosquitto_sub's retain flag; `%R` is the
*response topic*. With `%R` the field came back empty, every message parsed as a live
publish, and a watch against a **powered-off** PLC reported 80 of 80 republished and
**passed**. It now refuses any flag that is not `0` or `1` rather than guessing.

Proven in **both** directions, which is the part worth having. Against a
powered-down PLC it reported 0 of 80 and exited 1. Against the real download that
carries the fix it reported 80 of 80 and exited 0, with no structural finding and no
composer error — and the `exp_aft=3` NOTE that named the eight stale configs is gone,
because they are no longer stale.

### Confirmed on hardware

Downloaded to the bench and watched across the cold start. All three things that
had to change, changed:

| | before | after |
|:--|:--|:--|
| discovery configs republished | **0 of 80** | **80 of 80** |
| `exp_aft` on the eight RTD configs | `3` | **`180`** |
| `composed truncated json` in the log | every entity, every start | **none** |

The broker now carries 195 configs at `exp_aft: 0` (deliberate, unchanged) and 8 at
`180` — three heartbeats of a one-minute interval, which is what the code always
computed and could never publish. No structural finding on any of the 80.

## Documentation

- `FB_INPUT_TEMPERATURE_RTD_MQTT.md` — the block, both fault modes, and what
  neither can catch (a wrong sensor type that still tracks).
- `docs/AnalogInputs/UsingRTDSensors.md` — which module, wiring (a Pt1000 has no
  polarity; the 463 is 2-conductor), who has to configure what, and a commissioning
  table where **whether the value moves matters as much as what it says**. It also
  records the measured behaviour that contradicts the data-sheet reading the block
  started from: an unwired channel here reads `0`, not over-range.
- `MQTT_SelfWiring.md` gains it in the self-wiring list; `MQTT_General.md`'s GVL
  region regenerates with the new prefix.
- The `codesys-loop` skill gains one line it was wrong about: a hand-written
  candidate *can* file itself into a folder, if it carries a `projectstructure`
  block with no `ObjectId`s. That is how this block landed in `BASIC/` rather than
  at the project root, where there is no scripted way to move it from.

## Known wart, carried over

The cyclic call sits in `PRG_MAIN.READ_PUSHBUTTONS`, whose name is now half a lie.
An analog channel is a physical input and that action is the one the chart runs
every cycle to take inputs in, but it deserves an action of its own — and an action
has to be associated with a step in an SFC chart to be called at all, which a
script cannot do. The comment in the code says so.

Installation projects need a re-sync to pick the block up. Nothing there
instantiates it, so it would arrive as an uncompiled pool POU until a site wires
one.


docs/AnalogInputs/UsingRTDSensors.md
```diff
@@ -0,0 +1,463 @@
+## Reading a temperature sensor (Pt1000) with a WAGO RTD module
+
+### **Content**
+
+How to connect a resistance temperature sensor to a WAGO 750-series RTD input module, map its
+channel in CODESYS, and publish it with
+[`FB_INPUT_TEMPERATURE_RTD_MQTT`](../FunctionBlocks/FB_INPUT_TEMPERATURE_RTD_MQTT.md). The
+reference bench uses a **750-451** and a Pt1000 on `R1`; the same steps apply to its relatives.
+
+The three **750-463** modules this page was originally written around were faulty and have been
```
MichielVanwelsenaere: stay to the point, there's no value to the reader mentioning they were faulty

docs/AnalogInputs/UsingRTDSensors.md
```diff
@@ -0,0 +1,463 @@
...
+outlasts the module.
+
+### **Which module**
+
+| Module | What it is | Sensor |
+|:--|:--|:--|
+| [750-451](https://www.wago.com/global/i-o-systems/8-channel-analog-input/p/750-451) | **8**-channel analog input | RTD resistance sensors, 2-conductor — the bench's module, set for **Pt1000** |
+| [750-463](https://www.wago.com/global/i-o-systems/4-channel-analog-input/p/750-463) | 4-channel analog input, adjustable | **Pt1000** / RTD resistance sensors, 2-conductor |
+| 750-461 | 2-channel analog input | Pt100 / RTD, supports 3-conductor |
+| 750-467 | 2-channel analog input | 0–10 V — **not** for an RTD |
```
MichielVanwelsenaere: > | 750-467 | 2-channel analog input | 0–10 V — **not** for an RTD |

> :rotating_light: **A resistance sensor cannot go on a voltage input.** The 0–10 V module measures
> voltage and has no excitation current, so a Pt1000 on it reads nothing meaningful. Check the
> module number printed on the front before wiring anything.

all of the above has no value, stay to the point. Remove!

docs/AnalogInputs/UsingRTDSensors.md
```diff
@@ -0,0 +1,463 @@
...
+[what the 750-463s turned out to be](#what-the-750-463s-turned-out-to-be) — because the method
+outlasts the module.
+
+### **Which module**
+
+| Module | What it is | Sensor |
+|:--|:--|:--|
+| [750-451](https://www.wago.com/global/i-o-systems/8-channel-analog-input/p/750-451) | **8**-channel analog input | RTD resistance sensors, 2-conductor — the bench's module, set for **Pt1000** |
+| [750-463](https://www.wago.com/global/i-o-systems/4-channel-analog-input/p/750-463) | 4-channel analog input, adjustable | **Pt1000** / RTD resistance sensors, 2-conductor |
+| 750-461 | 2-channel analog input | Pt100 / RTD, supports 3-conductor |
```
MichielVanwelsenaere: I'm told there are 461 modules that support PT1000 as well, there are variants. Be more accurate.

docs/AnalogInputs/UsingRTDSensors.md
```diff
@@ -0,0 +1,463 @@
...
+  from a data sheet: a Pt1000 on `R1` reads 25.9 °C in a room that is about that. A Pt1000 on a
+  channel set for Pt100 sees ten times the resistance it expects and saturates at the top of
+  scale, so **a correct absolute reading is itself the proof of the sensor type** — no tooling
+  needed to establish it.
+- The **750-463 is the Pt1000 variant**, and Pt1000 is its factory default. If the module has
+  never been reconfigured, **a Pt1000 needs no configuration at all.**
+- It is *adjustable* — Pt1000, Ni1000, KTY81 and plain resistance ranges — and those settings live
+  in the module, changed with **WAGO-I/O-CHECK** over the controller's service interface, or by
+  register communication through the process image.
+- **Register communication is not available at all, and not because of how this project is set
```
MichielVanwelsenaere: this entire bullet has no value for reader, remove any reference to 'register communication'. Only describe Wago IO Check

docs/AnalogInputs/UsingRTDSensors.md
```diff
@@ -0,0 +1,463 @@
...
+  story one size up: `01C3_750_451` is a Modulecode, **eight** `INT` input channels and eight
+  length WORDs, and nothing else. There is no sensor-type parameter, no control/status byte and no
+  second variant to switch to, so there is nothing to enable in the IDE either. Its own description calls the module *"parameterizable"* — just not
+  from here.
+
+So: a factory-default 750-463 with a Pt1000 is plug-and-play. Anything else is a
+[WAGO-I/O-CHECK](../WagoIoCheck.md) job at a laptop, and the symptom that says so is described
+below.
+
+:rotating_light: **The bench's three 750-463 modules were broken, and configuration was never the
```
MichielVanwelsenaere: no value to user, remove

docs/AnalogInputs/UsingRTDSensors.md
```diff
@@ -0,0 +1,463 @@
...
+problem.** Every channel read exactly `1500` — 150.0 °C — and never moved a digit, with a Pt1000
+connected. This page spent a long time assuming that meant they had been set up for some other
+sensor. They had not: WAGO-I/O-CHECK eventually showed every parameter already correct and the
+analog front end producing no conversion at all. They have been taken off the rail; the 750-451
+that replaced them works. See [where that leaves it](#where-that-leaves-it), and
+[configuring a module with WAGO-I/O-CHECK](../WagoIoCheck.md) for how that was established.
+
+### **Mapping the channel in CODESYS**
```
MichielVanwelsenaere, line 94: this entire section seems oriented for a LLM, not a human reader, remove it

docs/AnalogInputs/UsingRTDSensors.md
```diff
@@ -0,0 +1,463 @@
...
+the tree, leaving `440, 540, 540_1, 463, 463, 463` on both sides — matched, verified — and all
+twelve channels still read exactly `1500`. So the tree order was a genuine defect worth fixing and
+it was not what made the channels read a constant. Both things were true at once, which is what
+made it convincing.
+
+:bulb: **Two faults at once is the normal case on a bench, not the exception.** Fixing one and
+re-testing is the only way to tell which one you were looking at; the mistake here was writing
+down a cause before re-testing.
+
+#### Where that leaves it
```
MichielVanwelsenaere: last comment on this, remove any reference or docs describing the faulty modules we experienced. No value to human reader.

docs/FunctionBlocks/FB_INPUT_TEMPERATURE_RTD_MQTT.md
```diff
@@ -0,0 +1,240 @@
...
+<!-- fb-badge:start -->
+![MQTT Discovery](https://img.shields.io/badge/MQTT%20Discovery-brightgreen)
+<!-- fb-badge:end -->
+
+### **General**
+
+Publishes a temperature read by an **RTD input module** — written for a Pt1000 on a WAGO
+[750-463](https://www.wago.com/global/i-o-systems/4-channel-analog-input/p/750-463), and running
+on the 8-channel
+[750-451](https://www.wago.com/global/i-o-systems/8-channel-analog-input/p/750-451) that replaced
```
MichielVanwelsenaere: > that replaced it

just mention on which modules it should work.

docs/FunctionBlocks/FB_INPUT_TEMPERATURE_RTD_MQTT.md
```diff
@@ -0,0 +1,288 @@
...
+
+### **Interface**
+
+**Inputs**
+
+| Pin | Type | Description |
+|:--|:--|:--|
+| `Raw` | INT | The mapped channel word of the RTD module, in tenths of a degree Celsius. Wire it to the channel variable — `Raw := RTD_001` — and nothing else is needed to make the block work. |
+| `LeadResistance` | REAL | Round-trip resistance of the sensor leads, in ohms; defaults to 0, which corrects nothing. The 750-451 and 750-463 both measure **2-conductor**, so the leads sit in series with the element and the channel reads *high* by them. A Pt1000 moves about 3.9 Ω/°C near room temperature, so the error is `LeadResistance / 3.9` °C — about 0.18 °C down 2×10 m of 0.5 mm², and 1.8 °C down 2×50 m of 0.25 mm². Measure it with the sensor disconnected, by shorting the far end of the pair and reading the loop at the terminals. :rotating_light: **Correct this in one place only** — see below. |
+| `PublishDeadband` | REAL | Degrees Celsius of change required before the value is published again. Defaults to 0.2. The last digit of an RTD channel jitters continuously, and nobody wants 0.1 °C of noise in their history; zero publishes every change. |
```
MichielVanwelsenaere: I don't see why we need this, HeartbeatInterval shoul be enough.

docs/FunctionBlocks/FB_INPUT_TEMPERATURE_RTD_MQTT.md
```diff
@@ -0,0 +1,288 @@
...
+
+**Outputs**
+
+| Pin | Type | Description |
+|:--|:--|:--|
+| `Temperature` | REAL | Degrees Celsius. **Held** at the last in-range reading while the channel is out of range, rather than following it into whatever the module reports for a broken wire. A *stuck* channel keeps updating it — the number may be right; what is in doubt is whether anything measured it. |
+| `Valid` | BOOL | The channel is reading a real sensor, within the range this sensor type can measure. |
+| `Fault` | BOOL | **Do not trust this reading.** Set when the value is outside what a platinum RTD can produce — usually a wire out of a terminal — and when the channel has stopped moving. This is what drives the entity's availability: while it is set, Home Assistant shows the sensor as unavailable. |
+| `Stuck` | BOOL | The channel has not moved a digit for `StaleTimeout`. Kept separate from `Fault` so a debugger can tell *impossible number* from *not measuring*; both set `Fault`. |
```
MichielVanwelsenaere: I think it's perfectly possible that the module measures the same temperature for 15 minutes if the checked value is 262 => 26,2 degrees for example. Am I missing something here?

docs/AnalogInputs/UsingRTDSensors.md
MichielVanwelsenaere: Removed. The intro no longer mentions them at all — it just says the bench uses a 750-451 with a Pt1000 on `R1`. (db47e59)

docs/AnalogInputs/UsingRTDSensors.md
MichielVanwelsenaere: Both gone — the 750-467 row and the "a resistance sensor cannot go on a voltage input" callout. (db47e59)

docs/AnalogInputs/UsingRTDSensors.md
MichielVanwelsenaere: You're right, and the table was wrong. Rebuilt from the device description this controller actually carries (`C:\ProgramData\CODESYS\Devices880 0001.19.0.0\device.xml`), which lists the variants explicitly:

- `75x-461/000-003` — **2AI, Pt 1000**
- `75x-461/000-006` — 2AI, Pt 100
- `75x-461/000-005` and `/000-009` — 2AI, Ni 1000
- plain `75x-461` — adjustable

So the row now says the ordering variant fixes the sensor. I also added the **750-450** (4 channels, 2-/3-/4-conductor), and a lead column, because `LeadResistance` depends on whether the module compensates. (db47e59)

docs/AnalogInputs/UsingRTDSensors.md
MichielVanwelsenaere: Gone, and so is every other mention of register communication — including the device-description reasoning about why it can't be reached. The section now says the sensor type lives inside the module and WAGO-I/O-CHECK is how you change it, and stops there. (db47e59)

docs/AnalogInputs/UsingRTDSensors.md
MichielVanwelsenaere: Removed. (db47e59)

docs/AnalogInputs/UsingRTDSensors.md
MichielVanwelsenaere: Trimmed to the one thing a human actually has to do: name each channel in the module's **I/O mapping** tab, plus the warning that the device-tree order must match the rail order. The mapped-variable table stays, since that is how you know which terminal feeds `RTD_001`.

Everything about the tooling is gone — the ScriptEngine's inability to place a module, the `-MapIo` spec, and the three-pass ordering note. Say the word if you want the remainder gone too; I kept it because a reader who skips it cannot get a channel to read at all. (db47e59)

docs/AnalogInputs/UsingRTDSensors.md
MichielVanwelsenaere: All of it is gone: the 148-line diagnosis with the two wrong theories, the "where that leaves it" and "the verdict" sections, and the worked example on the WAGO-I/O-CHECK page as well.

Across the three docs that is **334 lines removed, 81 added**. What survives about the 750-463 is only the module itself — its terminal-pair order (`R1, R3, R2, R4`) and its open-circuit value — which is still true of any 750-463 and useful to whoever has one. (db47e59)

docs/FunctionBlocks/FB_INPUT_TEMPERATURE_RTD_MQTT.md
MichielVanwelsenaere: Rewritten. It now says it works with any WAGO 750-series RTD module that puts tenths of a degree in the process image — **750-450, 750-451, 75x-461, 750-463** — and that the channel count and sensor type are the module's business, not the block's. No story about which replaced which. (db47e59)

docs/FunctionBlocks/FB_INPUT_TEMPERATURE_RTD_MQTT.md
MichielVanwelsenaere: Agreed, and it is now removed entirely — the `StaleTimeout` input, the `Stuck` output, the timer and the sample memory. `Fault` is simply `NOT Valid`.

To answer the "why did it exist" half: it was not the same question as `HeartbeatInterval`. The heartbeat plus `expire_after` detects a dead *publisher*; the stale check was meant to detect a live publisher reporting a dead *sensor*. But the only case it ever demonstrably caught — an open circuit parked at an end of scale — is now rejected on the first scan by `PlausibleMin`/`PlausibleMax`, which landed later and quietly took over its job. See the reply on your next comment for the measurement that settles it. (6174a33)

docs/FunctionBlocks/FB_INPUT_TEMPERATURE_RTD_MQTT.md
MichielVanwelsenaere: You are not missing anything — the doc was wrong, and this bench proves it.

`RTD_001` went `259 → 260 → 261 → 262` over about three and a half hours: roughly **one digit every 75 minutes**, sitting perfectly still in between. So a 15-minute window is inside a quiet period something like 80% of the time.

And the consequence was not cosmetic. `Fault := (NOT Valid) OR Stuck`, and `Fault` publishes `availability: offline` — so a false `Stuck` takes a **working** sensor out of Home Assistant. That is a worse failure than the one it was guarding against.

Removed entirely in 6174a33. Verified after a cold start, the last line being the point:

```
wired channel      Raw=262   26.2 C   Valid=TRUE   Fault=FALSE
unwired channel    Raw=8500  850.0 C  Valid=FALSE  Fault=TRUE
45 s later, same   Raw=262   26.2 C   Valid=TRUE   Fault=FALSE
```

The docs now name what is *not* guarded instead of implying the surface is complete: a plausible frozen value, and a wrong sensor type that still tracks. Both are commissioning problems — a thermometer, and one short across the terminals.

docs/FunctionBlocks/FB_INPUT_TEMPERATURE_RTD_MQTT.md
```diff
@@ -0,0 +1,291 @@
...
+published to `/TEMP`. The entity is briefly unavailable after a restart instead of briefly wrong.
+
+### **What is not guarded, and why**
+
+A channel can still lie in two ways this block does not catch, and it is better to name them than
+to imply otherwise.
+
+**A plausible number that is not a measurement.** A front end that has stopped converting can park
+the channel at a believable room temperature, and nothing in the process image says so. There was
+a movement check here — a channel that had not moved a digit for `StaleTimeout` was flagged
```
MichielVanwelsenaere: remove any reference to StaleTimeout, don't mention things that are not implemented/present.

docs/FunctionBlocks/FB_INPUT_TEMPERATURE_RTD_MQTT.md
```diff
@@ -0,0 +1,291 @@
...
+### **Home Assistant**
+
+The block publishes its own discovery config, so no YAML is needed. One entity appears, under the
+PLC's own device:
+
+| Entity | Category | `device_class` | `state_class` | Unit |
+|:--|:--|:--|:--|:--|
+| *FriendlyName* | — | `temperature` | `measurement` | °C |
+
+**There is deliberately no diagnostic entity.** An earlier version published a `problem` binary
```
MichielVanwelsenaere: same here

docs/FunctionBlocks/FB_INPUT_TEMPERATURE_RTD_MQTT.md
```diff
@@ -0,0 +1,291 @@
...
+
+```
+"avty": [ {"topic": "Devices/PLC/Lab/availability"},
+          {"topic": "Devices/PLC/Lab/Out/AnalogInputs/fbAiRtd001/availability"} ]
+```
+
+so the sensor is shown only while the PLC **and** the channel both say `online`. The second slot
+is free to use because a PLC discovery device fills both with the same topic, so overriding one
+loses nothing — see `CreateSensorEntity`'s `AvailabilityTopic`.
+
```
MichielVanwelsenaere, line 279: same here

docs/FunctionBlocks/FB_INPUT_TEMPERATURE_RTD_MQTT.md
MichielVanwelsenaere: Removed. The section was explaining why a check *isn't* there, which is itself documenting something that does not exist.

The **limitation** is real and stays, because it is a property of the block as it stands: a front end frozen at a believable temperature is not caught, and cannot be, since it is indistinguishable from a quiet room. The paragraph now says that and points at the test that does work — short the terminals and see whether the channel drops. No mention of what used to be there. (1c91de1)

docs/FunctionBlocks/FB_INPUT_TEMPERATURE_RTD_MQTT.md
MichielVanwelsenaere: Removed. "An earlier version published a `problem` binary sensor" explained the current design by contrast with one nobody can see.

It now simply states what is true: the fault is carried by availability rather than by a separate entity, so it greys out the temperature where somebody is already reading it instead of hiding in the diagnostics section of the device page. (1c91de1)

docs/FunctionBlocks/FB_INPUT_TEMPERATURE_RTD_MQTT.md
MichielVanwelsenaere: Removed — that callout was migration advice for upgrading from a version that is not in this project.

While applying these three I found a fourth of the same kind and removed it too: `DataAvailable`'s description still said *"it used to latch on the zero a mapped channel reads before the K-bus fills the process image; the startup gate below fixed that"*. It now just says what it does. (1c91de1)

PR #195, opened 2026-08-30
Announce the pushbutton dimmer to Home Assistant: a button and a level

MichielVanwelsenaere:
`FB_INPUT_PUSHBUTTON_DIMMER_MQTT` was the last input block with no MQTT discovery.
`MQTT_SelfWiring.md` already named it as the worthwhile follow-up; this does it.

It already extended `FB_MQTT_BASE` — the comment in `PRG_MAIN` saying it did not was
stale — so all it lacked was an `InitMqttDiscovery` method and the self-wiring prologue.

## Which entities, and why

The block is two things on one topic tree, so it announces two:

| Entity | Platform | State topic | Behaviour |
|:--|:--|:--|:--|
| *FriendlyName* | `event`, `dev_cla: button` | `…/<block>` | Fires on each press. Types `SINGLE`, `DOUBLE`, `LONG`. Stateless, no expiry. |
| *FriendlyName* level | `sensor` | `…/<block>/DIM` | The dim level, 0-255. No unit, no device class, no state class. `expire_after: 0`. |

Both belong to the PLC's discovery device and inherit its availability. Neither carries an
availability topic of its own: nothing here can be individually untrustworthy the way an
RTD channel can.

`Q`, `DBL` and `P_LONG` keep publishing but are **not** announced. `P_LONG` is the hold
duration of a `LONG` the event entity already carries, and `DBL` is internal toggle state.

**The level is a sensor and not a light, deliberately.** A light entity has to be
commandable, and this block has no subscribe side at all — no callback collector, no
command topics. `FB_OUTPUT_DIMMER_MQTT` is the block that has those, and stays the answer
when Home Assistant must *set* the level rather than watch it. Growing this one into a
light would be a second implementation of the block next door.

`expire_after` is 0 because there is no heartbeat here: `DIM` publishes on change, gated by
`Delta_Dimm`, so any positive expiry would grey out a level that is merely steady.

## Behavioural changes

:warning: **The three press payloads are now JSON.** They were the bare words `SINGLE`,
`DOUBLE` and `LONG`. Home Assistant's event platform reads `event_type` out of the payload,
so a bare word arrives and is discarded — the same change `FB_INPUT_PUSHBUTTON_MQTT` made
when it gained discovery. Anything still consuming the old payload, hand-written YAML
included, has to follow.

**`DIM` publishes retained, and once at startup.** It is state, not an event: a Home
Assistant restart should not leave the sensor *unknown*, and a level that has never moved
would otherwise never be published at all, leaving a freshly discovered entity empty
forever.

**`DIM` uses `Qos_Dimm`.** It was hardcoded to `ExactlyOnce` while `Qos_Dimm` was being
applied to `Q` instead — the reverse of what the pin name and the docs page have always
said. No practical change at the lab, which passes `ExactlyOnce` anyway.

`fbDiPb002` loses its explicit `InitMqtt` call and takes a `FriendlyName`.

## A defect this exposed

`CreateSensorEntity` declared `DeviceClass: STRING := 'None'`, and its body only publishes
`null` when the value is `''`. Home Assistant validates `dev_cla` against a fixed list and
**rejects the whole config** when it does not recognise the value, so the entity never
appears at all — the failure mode the SDM220's `reactive_energy` comment already describes.

Every existing caller passes a real class, so that default had never been exercised. The
dim level is the first entity that legitimately has none, and its first published config
carried `"dev_cla":"None"`. The default is now `''`.

`check_mqtt_discovery.py` gained a guard: a `dev_cla` that is the string `'None'`, `'null'`
or empty is a failure. Validated the right way round — run against the broken config still
on the broker it failed, naming the topic; after the fix, clean.

## Verification

Compile-verified against the recorded baseline: `NEW vs baseline: 0`, the usual 9 warnings
and no new ones. `structValue` in the export confirms the `FriendlyName` initialiser
actually reached the compiler, not just the declaration text.

Downloaded to the lab PFC200 and read off the broker:

- both configs published and structurally valid, `dev_cla: null` on the sensor
- `…/fbDiPb002/DIM` retained at `0` — the startup publish works
- `check_mqtt_discovery.py`: 82 configs, `OK`
- `check_string_widths.py`: `OK`, 19 wide strings examined
- `update-fb-docs --check`: clean

**Not yet proven on hardware: the level moving.** That needs a finger on `DI_002` — a press
and a hold — and the watch window closed with nobody at the bench. The event entity, the
sensor config and the startup value are all confirmed; the ramp is not.

https://claude.ai/code/session_01EAikZTFhnTLFc8v3u3T8zv


docs/AdditionalFunctionality/MQTT_SelfWiring.md
```diff
@@ -159,16 +159,19 @@ a block's first body call — so self-wiring them while leaving discovery in the
...
-are the worked examples of what that looks like for an RS485 block. Until then their
-call sites are untouched and keep working exactly as before.
+**Not self-wired.** `FB_RS485_DUCO_DUCOBOX_MQTT` *does* extend `FB_MQTT_BASE`, so
+`FriendlyName` is already on it. What it lacks is an `InitMqttDiscovery` method — a
+prologue could wire its MQTT publishing but would have nothing to announce. Giving it
+discovery is the worthwhile follow-up, and the three Eastron meter blocks are the
+worked examples of what that looks like for an RS485 block. Until then its call site
+is untouched and keeps working exactly as before.
+
+`FB_INPUT_PUSHBUTTON_DIMMER_MQTT` was the other one on this list. It now announces an
```
MichielVanwelsenaere: Never mention history on docs pages. Only current state. Create a Claude.md file for this in the docs directory and take note of this for future sessions

docs/FunctionBlocks/FB_INPUT_PUSHBUTTON_DIMMER_MQTT.md
```diff
@@ -72,94 +77,116 @@ BOOL ──┤ RST                      SINGLE ├── BOOL
...
 | **Output changes: Q**   | A change is detected on output `Q`. (*) | `TRUE/FALSE` | 2 | `TRUE` | no
 | **Output changes: DBL**   | A change is detected on output `DBL`. (*) | `TRUE/FALSE` | 2 | `TRUE` | no
-| **Output changes: DIM**   | A change is detected on output `DIM`. (*) | `0-255` | configured in method call `InitMQTT` | `FALSE` | no
+| **Output changes: DIM**   | The level has moved by at least `Delta_Dimm`, or the button was released. (*) | `0-255` | `Qos_Dimm`, from `InitMQTT` | `TRUE` | **yes**
 
 MQTT publish topic is a concatenation of the publish prefix variable and the function block name.
 
 (*): MQTT publish topic is a concatenation of the publish prefix variable, the function block name and the name of the output. 
 
+:warning: **The three press payloads are JSON.** They used to be the bare words `SINGLE`,
```
MichielVanwelsenaere: Again: no history

docs/AdditionalFunctionality/MQTT_SelfWiring.md
MichielVanwelsenaere: Done, and written down.

The paragraph is gone. It was redundant anyway — the block is named in the **Self-wiring** list above, which is the current-state way to say the same thing, and the JSON payload fact lives on the block's own page where a reader needs it.

I went further on this file than the one line, because three older passages broke the same rule and leaving them would have the file contradicting the rule it was given about: the opening framed as *blocks used to need two calls*, *everything it used to be handed*, and *behaves exactly as it did before this change*. All reframed from the present, same facts. Say the word if you would rather I had left those alone.

The rule is now `docs/CLAUDE.md`, which loads automatically when editing under `docs/`. It names the phrasings that give it away — including *now announces*, which reads as present tense but is a comparison with a past the reader cannot see — says where the history goes instead (commit message, PR description, `git blame`), and marks out what it does **not** forbid: rationale, naming a limitation, pointing at the block to use instead, and genuinely dated facts like a device-description version. The root `CLAUDE.md` points at it. Saved to my memory for future sessions too.

One thing I did **not** strip, so it is a decision and not an oversight: two comments in the ST still explain a trap historically — `CreateSensorEntity` saying its `DeviceClass` used to default to `'None'`, and the dimmer body noting `Qos_Dimm` was being applied to `Q`. Those are at the code, where they stop someone changing it back, which is the exception `docs/CLAUDE.md` allows. Happy to strip them if you want the rule to reach code comments too.

docs/FunctionBlocks/FB_INPUT_PUSHBUTTON_DIMMER_MQTT.md
MichielVanwelsenaere: Rewritten. It now states what the payloads are and why, with no account of what they replaced:

> :bulb: **The three press payloads are JSON.** Home Assistant's event platform reads `event_type` out of the payload, so a bare word would arrive and be discarded. Anything reading these topics directly, hand-written YAML included, has to read the JSON too.

The migration warning that was there — that anything on the old bare-word payload has to follow — is in the PR description instead, which is what gets read at merge time. The page keeps the consequence for a reader who has never seen the old payload: read the JSON.

Downgraded from `:warning:` to `:bulb:` for the same reason: the warning was about the change, not about the behaviour.

PR #197, opened 2026-09-01
Speak to a DucoBox Focus, and give each publishing task its own queue lane

MichielVanwelsenaere:
Builds out the DucoBox Focus block into something Home Assistant can both read and drive, and fixes a publish-queue race that the extra traffic exposed.

## The queue race

`FB_MQTT_PUBLISH_QUEUE` was single-producer by contract and multi-producer in fact. Its own comment named the failure: two tasks inside `AddMessage` at the same time read the same write index, write into the same slot, and leave a hybrid there — one message's topic carrying another's payload.

Measured on a live controller: a device published `offline` it had never sent — three availability messages by its own count against seventeen on the broker, while the bus reported `ok=747 fail=1`. Growing the RS485 side turned a latent race into a constant one.

The ring is now a lane per producing task, drained round-robin by the single consumer. Each lane is genuinely single-producer/single-consumer: its write index is touched only by the owning task, its read index only by the drain, and each side merely reads the other's index — one aligned `INT` load on this 32-bit target, which cannot tear. No semaphore, so no priority inversion. Lanes split the existing storage rather than adding any; `ST_MQTT_MESSAGE` is ~1.8 kB, so a second queue instance was never affordable.

`DroppedCount` had been incremented since the queue was written and never published. It now rides in the RS485 diagnostic as `drop=`.

## The Duco block

- Renamed to name the model it speaks to; the master register block is model-specific.
- Decodes the components instead of publishing raw registers, resolving each one's type first — on a window ventilator and an actuator board the same register means different things.
- Announces each component as its own Home Assistant device, with the entity set its type actually has.
- **Write path**: numbers, buttons and switches command `<sub>/<node>/write/<register>`; the write is paired with a read-back of the same register under `AbortOnError`, so what gets published is what the device holds rather than what it was told. The box itself is node 1 and is written the same way.
- **Settings read-back**: every announced control now gets a value before anybody sets one. One register per transaction, at POLL level, so refresh traffic can never crowd out a command or a measurement.
- Values publish on change with a five-minute refresh. Before that, six unit topics sent 126 identical messages per 100 s.
- **Self-wiring**, and it shows what that means for a block that speaks for more than itself. `FriendlyName` names the box; each component keeps the name its `AddNode` call gave it and is announced once the bus has reported what type it is. So the prologue wires the box, and the components follow from the bus rather than from the call site.

## Discovery

`CreateNumberEntity` and `CreateButtonEntity` are new here, and both assigned `dev_cla` unconditionally. Home Assistant validates it against a closed list and rejects the **whole** config when it does not match, so an empty string did not leave a button unclassified — it stopped the button existing:

```
expected ButtonDeviceClass or one of 'identify', 'restart', 'update'
for dictionary value @ data['device_class']
```

`CreateSensorEntity` had guarded this since it was written. Both now do the same, for the entity category and the number's unit as well.

## Naming the temporary settings

Holding 9 actions 2, 3 and 4 are *temporary* manual settings: the zone returns to automatic on its own. Nothing said so — the status the box reports back is a plain `Manual low`, with no hint that it will lapse — so the buttons carry it in their names.

They say `(timed)`, not a duration. L8000009 gives 15 minutes on an extraction system and 8 hours on a supply one, and the figure is not fixed either: a room operating unit holds it in holding 7, settable from 5 to 9995 minutes, which this block already announces as that component's `manual time` control. A number in the button name would have been a default presented as a fact, changeable from the same dashboard it appeared on, and wrong outright on a supply-linked node.

The 10, 20 and 30 minute high settings cannot be requested at all. They exist only as `INPUT 1` status values, so a physical control can produce one and this block reports it, but no action value asks for one. `target`, holding 0, is the other way to overrule — a level with no timer.

## Also here

- A binary sensor can be held at rest until the K-bus is really exchanging data. `xConfigFinished` goes true before the first input words arrive, and through inverted normally-closed loops a zero-filled image reads as alarm.
- `resolve_node` addresses a device by full path, so `-Under Pfc200Bus` is no longer a coin flip on a project with two controllers.
- RS485 pages no longer record what has been run against them; a page describes the block, not its test history.
- An input's default value now shows on its own page — 32 inputs across 16 pages had one the generator knew and never printed.
- The DucoBox page described the wiring it had before it self-wired, and the binary sensor page had a code sample that would not compile — it called the function block type rather than an instance, and named a global variable list that no longer exists.

## Verified

On hardware: the decode, the discovery configs, availability (`drop=0`, no spurious offline), publish-on-change, and a write to the box.

Not on hardware: writes to the components, and the settings read-back. Both use the same function-3 pairing the box write proved.

The Duco Modbus interface needs **revision 1** of the DUCO communication board; later revisions do not offer Modbus RTU at all. The page says so.

https://claude.ai/code/session_01EAikZTFhnTLFc8v3u3T8zv


docs/AdditionalFunctionality/MQTT_SelfWiring.md
MichielVanwelsenaere, line 1: It is self wired now? If not, make it so.

docs/FunctionBlocks/FB_INPUT_BINARYSENSOR_MQTT.md
```diff
@@ -10,23 +10,24 @@ Binary sensors gather information about the state of devices which have a "digit
...
 
 **Inputs**
 
-| Pin | Type | Description |
-|:--|:--|:--|
-| `BS` | BOOL | Digital input linked to the signal wire of the binary sensor. |
+| Pin | Type | Default | Description |
+|:--|:--|:--|:--|
+| `BS` | BOOL |  | Digital input linked to the signal wire of the binary sensor. |
+| `ProcessImageValid` | BOOL | `TRUE` | TRUE once the K-bus has finished starting and the mapped inputs mean something; FALSE holds the sensor at rest. Wire it to the bus driver's own answer — `Pfc200Bus.xConfigFinished`. Defaults to TRUE, so a call site that says nothing behaves as before. **Pass it on any loop that is normally closed**: until the bus is up every mapped word reads zero, and through the inversion such a loop needs, zero *is* the tripped state — so an ungated detector raises its alarm on every cold start. It gates the debounce, which still has to settle afterwards, so a genuine alarm present at power-up is reported a few cycles late rather than not at all. |
```
MichielVanwelsenaere: Way to much text. Keep it to the point. Make sure the code sample shows something that works.

docs/AdditionalFunctionality/MQTT_SelfWiring.md
MichielVanwelsenaere: Done in 672c2c5 — the box self-wires now. FriendlyName names the box; each component keeps the name its AddNode call gave it and is announced once the bus reports its type, so the prologue wires the box and the components follow from the bus. The page's claim that it lacks InitMqttDiscovery was stale — this PR adds that method — so I rewrote the paragraph and added it to the summary list.

docs/FunctionBlocks/FB_INPUT_BINARYSENSOR_MQTT.md
MichielVanwelsenaere: Done in 672c2c5. The ProcessImageValid cell is down to two sentences. The sample was worse than verbose — it called the function block *type* instead of an instance, and named MQTTVariables, which no longer exists. Replaced with the self-wiring form: a declaration carrying FriendlyName, the cyclic call with the NOT and the ProcessImageValid gate, and one edge-driven output. Page is 104 lines, down from 117.

PR #198, opened 2026-09-03
Fix a review's findings, and stop the docs describing a queue that is gone

MichielVanwelsenaere:
Sixteen review findings, triaged against the code before anything was touched. Twelve were real and are fixed here; one was a false positive; three are real and deliberately left, with the reasons below.

Every fix is compile-checked in both applications, `NEW vs baseline: 0`, nine warnings unchanged. `FB_OUTPUT_DIMMER_DMX_MQTT` has no instance anywhere so it is never normally compiled — it was verified under the `verify` harness, which instantiated it: 0 errors.

## Verdicts

| # | Finding | Verdict |
|:--|:--|:--|
| 1 | `F_STRIP_JSON_ROOT` unguarded negative length | fixed — reachability narrower than reported |
| 2 | Valve state payloads inverted | fixed |
| 3 | `Judge` `LastFailBytes` loop pasted twice | fixed |
| 4 | Thermostat hysteresis permanently zero | fixed |
| 5 | DMX channel indexing off by two | fixed |
| 6 | `DmxWidth` / `DmxUniverse` stored and ignored | partly fixed, rest documented |
| 7 | Esera temperatures cannot go below zero | fixed |
| 8 | `sInstanceName` read four times, never written | **false positive** |
| 9 | Ping task blocks at priority 4 | fixed |
| 10 | Publish queue costs ~1.7 MB | real, not attempted |
| 11 | Dead locals in every `Create*Entity` | fixed — 16 methods, not 14 |
| 12 | `PublishEntityConfig` hardcodes `DINT#1499` | fixed |
| 13 | `PRG_MQTT` worker count declared twice | fixed |
| 14 | `Init` does not validate its transport | fixed, both methods |
| 15 | Multi-producer race on the queue | mostly a **stale docs page**; one real residual fixed |
| 16 | FB-level `VAR PERSISTENT` probably not persisting | confirmed by the compiler, needs the IDE |

## The two that were silent

**A valve entity read exactly backwards.** `InitMqttDiscoveryAsValve` paired `PayloadOpen := (NO)` with `StateOpen := (NC)`, where every sibling method pairs like with like — and `RelayType`'s own doc comment says it "swaps the payload pair so the entity reports the state of the load". Commanding open wrote TRUE, the PLC published TRUE, and Home Assistant read TRUE as closed.

**A thermostat had no deadband.** `Hysteresis : REAL;` had no initialiser and `ConfigureFunctionBlock` is called on no thermostat in the project — only on `fbAoDimmer001`. At 0.0, heat mode reduces to `temp < setpoint → ON`, `temp > setpoint → OFF`, on a 50 ms task, each flip a retained QoS 2 publish. It defaults to `0.5` now, matching the `TempStep` the climate entity is announced with.

## Finding 15 was a documentation bug

The race was fixed in #197, which gave each producing task its own lane. `docs/AdditionalFunctionality/MQTT_General.md` was never updated and still described "a ring buffer of 1025 slots" under a heading reading **What is still not safe** — so the page described the past as the present, which is what the review read. That page now describes lanes, the lane table, and the cost of adding one.

Two things really were still wrong underneath it:

- **HvacTask published on lane 0**, MainTask's own — and the queue's own header says in as many words that sharing a lane between tasks "brings it straight back". Its four blocks name their own lane now.
- **Lanes were picked by number**, with nothing recording which task owned which. `GVL_MQTT.MQTT_LANE_MAIN`, `_RS485` and `_HVAC` name them, and the RS485 blocks' 18 literal `1`s are gone.

A lane is not free: a slot holds a `STRING(1500)` payload plus a `STRING(160)` topic, about 1.8 kB, so the three lanes hold **341 slots each** rather than 512 and total storage stays at the 1024 the single ring had. `DroppedCount` is published in the RS485 diagnostic as `drop=`, so if 340 per lane turns out to be tight it is visible rather than silent.

**One case is left, and the page says so.** `GVL_MQTT.PLC_Device` is announced to from MainTask's blocks *and* from the HVAC blocks, so its lane cannot be made per-task without moving one task's discovery onto the other's. Its publishes stay on lane 0. The window is startup, when discovery configs go out; steady-state publishing is on the right lanes.

## DMX: two halves, two conventions

The block writes `BUFFER[DmxChannel - 1]`; `PRG_DMX_SEND` read `BUFFER[iI + 2]`. Net effect: channel `c` lit slot `c - 2`, and channels 1 and 2 could not be reached at all, because no loop iteration ever read buffer index 0 or 1. The loop carried the author's own `// iI+2 is strange ... todo`.

Both halves now agree that **`BUFFER[c - 1]` is channel `c`**, and the convention is stated where the copy happens and on the page. Nothing in either project instantiates the DMX block, so no fixture is currently addressed against the old offset — but a fixture that was would need re-addressing.

`initDMX` also refuses a channel outside 1..512 instead of writing `BUFFER[-1]` for a 0. `DmxWidth` and `DmxUniverse` are still not acted on; that is a feature, not a fix, so the page now has a table saying exactly what each input does and does not do.

## Also here: the `meta` property is gone

Not from the review — asked for separately, and it turned out to be the same shape as finding 11. Every discovery config on the broker carried `"meta":null`, 44 of them in the last snapshot, for a property nothing has ever set.

It was plumbed the whole way through: an input on 24 block `InitMqttDiscovery*` methods, forwarded through 19 `Create*Entity` methods into `FillCommonFields`, and a member of `MQTT_DISCOVERY_BASE` — which is what put the key in the payload. `FillCommonFields` assigned it only when non-empty:

```iecst
IF NOT StrEquals(str1 := ADR(Meta), str2 := ADR('')) THEN
    pBase^.meta.CharString := Meta;
END_IF
```

and it never was. All 24 block inputs default to `''`, no call site passes anything else, and no program mentions `meta` at all — so the assignment never ran and the struct member serialised as `null` regardless. **44 declarations and 140 call arguments**, all dead.

The docs called it "Extra JSON merged into the discovery config", which is not what the code did either — it assigned a plain string to a `JSONVAR` and merged nothing. Those rows are generated from the code, so they went with it.

Home Assistant tolerated the key, so no entity changes. It just stops every config carrying 13 bytes that mean nothing, in a payload whose `STRING(1500)` truncation guard exists because the space is not free.

## Finding 8 is a false positive

`sInstanceName` is not unassigned. It carries `{attribute 'instance-path'}` with `{attribute 'noinit'}`, so the compiler fills it in, and the broker snapshots under `.ai/mqtt/` show the full path in every discovery log line:

```
.../diagnostic/Log  Wago_PFC200_G1_Lab.HomeAutomation.GVL_RS485.FB_RS485_EASTRON_SDM220_1.fbMqttDisc | Added ..._TOTRE as MQTT discovery
```

Same class as the reviewer's own `TO_STRING(eHvacMode)` note.

## Left undone, on purpose

**The queue's ~1.7 MB (10).** Real, and the numbers are now written down where they belong. A two-tier queue is a rewrite of the hottest path in the project and wants its own change.

**Persistence (16).** Confirmed without hardware — the compiler says it outright: `No matching instance path in VAR_PERSISTENT list found for variable PRG_MAIN.fbDoSwAlarmFire.bPreviousState`, and `GVL_PERSISTENT` is empty in the export. Registering the instance paths is the IDE's *Add all instance paths* command; no script can reach it. Worth doing, and worth asking afterwards whether a value that must survive a power cycle belongs in an FB instance at all rather than in `GVL_PERSISTENT` directly.

**Naming.** Left alone, as the reviewer did — `src/CLAUDE.md` says it is settled. The one exception is `udiIpublishers`, which became `cudiPublishers` because it changed kind to `VAR CONSTANT`.

## Not verified on hardware

Nothing here has run on a PLC. The compiler covers the type errors and the dead code; it cannot see the valve inversion, the thermostat deadband, the DMX offset or the Esera sign. Those four want a bench check, and the DMX and Esera paths want the hardware they name.

The `meta` removal is the one change whose effect is visible without a PLC: after the next download, no published discovery config should contain `"meta"`, and every entity should still be there. `py tools/ai/check_mqtt_discovery.py` reads the broker for exactly that.

https://claude.ai/code/session_01RDLJqGQXPxDu6daU4iC2wN


docs/AdditionalFunctionality/DMX_artnet.md
```diff
@@ -15,6 +15,28 @@ In `GVL_DMX` you can set an IP. This depends on your topology. Multicast works i
 
 In `PRG_DMX_SEND` you can set the universe. `0` is not recommended for Art-Net, so the default is `1`.
 
+<ins>Channel numbering</ins></br>
```
MichielVanwelsenaere: what's the point of this header style? has a strange effect on generated markdown, shows as regular text, not a header

docs/AdditionalFunctionality/DMX_artnet.md
```diff
@@ -15,6 +15,28 @@ In `GVL_DMX` you can set an IP. This depends on your topology. Multicast works i
...
+One convention, and both halves are written to it: **`GVL_DMX.DMX.BUFFER[c - 1]` is
+DMX channel `c`.** `FB_OUTPUT_DIMMER_DMX_MQTT` writes that slot, and `PRG_DMX_SEND`
+copies buffer index `i` into `stSBuf1.BUFFER[i + 18]` — an ArtDmx packet carries its
+18-byte header first and then channel 1 at data offset 0, so nothing shifts along
+the way. A channel outside 1..512 is refused by `initDMX`, which leaves the block
+dormant rather than indexing outside the buffer.
+
+<ins>What a block declares and what actually happens</ins></br>
```
MichielVanwelsenaere: same here

docs/AdditionalFunctionality/DMX_artnet.md
```diff
@@ -15,6 +15,28 @@ In `GVL_DMX` you can set an IP. This depends on your topology. Multicast works i
 
 In `PRG_DMX_SEND` you can set the universe. `0` is not recommended for Art-Net, so the default is `1`.
 
+<ins>Channel numbering</ins></br>
+One convention, and both halves are written to it: **`GVL_DMX.DMX.BUFFER[c - 1]` is
```
MichielVanwelsenaere: this entire section is quite technical, does it really matter for the end user?

docs/AdditionalFunctionality/MQTT_General.md
```diff
@@ -53,23 +53,53 @@ VAR_GLOBAL
...
 
 | Member | Kind | Meaning |
 |:--|:--|:--|
 | `HasMessage()` | METHOD : BOOL | Something is queued. This is what the drain loop gates on. |
 | `IsFull()` | METHOD : BOOL | No room for another message. Reports full one slot early. |
-| `DroppedCount` | VAR_OUTPUT : UDINT | Messages discarded because the ring was full. Should stay at 0. |
+| `DroppedCount` | VAR_OUTPUT : UDINT | Messages discarded because a lane was full. Should stay at 0. It rides in the RS485 diagnostic as `drop=`. |
 
 `HasMessage()` and `IsFull()` are **methods, not flags** — call them, do not read a
 stored value. There used to be `EMPTY` and `FULL` outputs and removing them was the
```
MichielVanwelsenaere: 'there used to be...', no history in the docs!

docs/AdditionalFunctionality/DMX_artnet.md
MichielVanwelsenaere: Converted to a real `#### **Channel numbering**`. I had copied the style from `<ins>Global parameters</ins>` further up the same page, so it was not new here — but you are right that it renders as ordinary text and gives the generated markdown no structure. I converted that pre-existing one too, since a half-converted page reads worse than either choice.

Left alone: `<ins>Birth message:</ins>` and `<ins>LWT message:</ins>` in MQTT_General.md. Those are captions on a topic/payload pair rather than section headings, so ordinary text is what they should be. Say the word if you want them changed as well.

docs/AdditionalFunctionality/DMX_artnet.md
MichielVanwelsenaere: Same fix — a real heading, and renamed to `#### **Metadata-only inputs**`, which says what the section is about in three words instead of eight.

docs/AdditionalFunctionality/DMX_artnet.md
MichielVanwelsenaere: Agreed, and cut. It explained how the block's buffer index reaches an ArtDmx slot, which nobody addressing a fixture needs to know. What is left is the part they do: `DmxChannel` is the channel that lights, the range is 1..512, and anything outside it is refused. The mechanism now lives only where it is enforced, in the `PRG_DMX_SEND` comment.

I kept the `DmxWidth` / `DmxUniverse` table, tightened. That one is not internals: both are accepted, published to Home Assistant, and then ignored, so setting `DmxWidth := 3` on an RGB fixture looks like it should work and does nothing. That is worth a user's time.

docs/AdditionalFunctionality/MQTT_General.md
MichielVanwelsenaere: Fixed, and it was two violations rather than one.

That paragraph predates this branch (#175), but the rule is the rule — it now says why a stored flag cannot be safe here, in the present tense, instead of what was removed. The consequences it described are still real, so they stayed; only the "there used to be" framing went.

I then checked my own additions to the same page and found the same fault: "On a live controller it showed up as a device publishing offline it had never sent, three availability messages against seventeen on the broker." That is a dated incident, not a description. It now says what reaches the broker when the race is hit, which is what a reader needs.

